Issue #11, opened 2019-11-26
Document potential research sources


spy-t:
As stated in the README this is an educational project and although it is not intended to replace the relevant course or textbook I think it would be wise documenting potential sources for anyone interested in learning about compilers in general and specifically about helping with the project.

baziotis:
Thanks mate, good one! And you just reminded me that I should make a new label.
Here are some resources I like: 
_Note_: I don't like much theory and classical textbooks for front-end code (you might not know
what front-end is - just skip this note, not so important). I really think that one will learn
better by practicing and by reading code of an actual compiler.
That said, I'm a big fan of theory when it comes to back-ends and _especially_ optimization!

- [Nora Sandler's "Writing a C compiler series"](https://norasandler.com/archive/)
Of course, you won't write a C compiler. :P
You will write a small subset of it which is nonetheless interesting. It inspired my first compiler
and using this subset one can implement first assignments in the course Introduction to Programming in UoA.
What is more, it takes an incremental approach. That is, it adds one feature at a time, by
teaching how to handle that feature in all parts of the compiler (lexer, parser, codegen).
In other words, it doesn't do: First let's build the _whole_ lexer, then the _whole_ parser etc.
That is good because first of all I think it's more educational and second, you can have an executable
from your _own_ compiler in an hour or 2 (it might only return 0 but hey, we all started somewhere).

- [Engineering a Compiler, 2nd Edition](https://www.elsevier.com/books/engineering-a-compiler/cooper/978-0-12-088478-0)
Truly a masterpiece. An incredible introduction to pretty much every topic (apart from lexing parsing
that I don't really like). Anything from chapter 4 and on is amazing IMO. I don't think it's
good to read it in one sitting but rather returning back to it from time to time.

- [Practical Compiler Construction](https://www.t3x.org/reload/)
Not really popular but I think it's a good one. The code has bad decisions in it but nonetheless,
it shows you how to implement an interesting subset of C.

Moving to more real-world resources:

- [9cc C compiler](https://github.com/rui314/9cc)
Actual C compiler with an educational point of view. I think it's a good starting point.

- [lacc C compiler](https://github.com/larmel/lacc)
Another actual C compiler with a custom IR. I like this one even more than 9cc I think.
Both are great though!

- [Bitwise](https://www.youtube.com/channel/UCguWV1bZg1QiWbY32vGnOLw)
This dude is a mastermind of code architecture. It's almost funny how he has achieved
[...]

PR #12, opened 2019-11-26
Create SOURCES.md

spy-t:
#11

### DIFF
--- a/SOURCES.md
+++ b/SOURCES.md
@@ -1,0 +1,4 @@
+# Research sources
+This is a list of sources about compilers and compiler development in general
+
+- [Awesome compilers](https://github.com/aalhour/awesome-compilers)


baziotis:
Thanks for the PR! Unfortunately I'm not sure about this one. Maybe just keeping [a discussion thread](https://github.com/ACM-UoA-Student-Chapter/dil/issues/11) is better than listing the resources into a file. This because probably
people will disagree on what's the best resource to learn from. I'm leaving this open though
if at some point we arrive at a list of resources that we all think is good and _also_ keep the discussion thread for other resources.

spy-t:
Noted. Although I think sources chosen by people like you who have previous experience with the subject can't go wrong, I agree that an open discussion might be better in the long run.

baziotis:
I partially agree but also people like me don't **necessarily** have useful opinions. :)
Also, experienced programmers might disagree with each other.
With the discussion, if one cares about anyone's specific opinions, they can read their posts and it's a win-win.

Issue #7, opened 2019-11-25
DIL Grammar


baziotis:
Grammar construction is one thing that I assume most people are not experienced with making.
However, I think it's worth it for you to try it. You can check the samples to try to understand the grammar, then try to construct the [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form).

To separate the effort, here's a list that I can hope I can make more detailed:
- [x] Expressions
- [x] Statements
- [x] Functions
- [x] Modules

DIL will follow an LLVM-like structure (as many other languages). That is, the top-level entity
is a module, which you can think as a file. A module has 0 or more functions.
A function has 0 or more statements. And a statement can use / contain expressions.

_Note:_ Αn important difference with e.g. C / C++ is that expressions are not intermingled with
statements and vice versa. In C, the assignment is both a statement and an expression, e.g.
you can do this: `a = b = c`; where `b = c` acts both as a statement and an expression.

If you're interested in this issue but you have no experience in formal grammars, we can incrementally build the grammar on Slack, starting from very easy grammars.

Doryanni:
Since this is a project supposed to be accessible by most students with some code development under their belts, should the grammar be LL(1) or should we aim for something different?

baziotis:
Right, LL(1). Generally, even in programming languages for which the compilers are developed from pretty experienced people, we aim for LL(1) grammars. That is because they're easier to lex / parse. Easier code means more maintainable and bug-free code. That said, LL(2) is also acceptable.

Doryanni:
Oh,  since I've already begun, may I be assigned to this issue?

baziotis:
Of course!

baziotis:
We completed a first draft with this: https://github.com/ACM-UoA-Student-Chapter/dil/pull/21
Closing for now.

PR #13, opened 2019-11-29
Began writing the grammar.

Doryanni:
Function definition, expression, and statements in progress.
I propose we make sub-branches from the grammar branch, one for each of the checklist's items.

### DIFF
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -1,0 +1,53 @@
+module = { [functionDef] }, "EOF";
+
+functionDef = "func", identifier, "(",  [funcDefArgList], ")", "{",  { [statement] }, "}";
+
+
+funcDefArgList = identifierList, ":", Type, funcDefArgListT;
+
+funcDefArgListT = ",", identifierList, ":", Type, funcDefArgListT
+                | ;
+
+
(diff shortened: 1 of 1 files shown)


baziotis:
Thanks for the PR!
> I propose we make sub-branches from the grammar branch, one for each of the checklist's items.

No need to do that and I think that it will complicate things. The grammar, as any multi-part issue, can be separated into multiple PRs.
Here are some notes I hope will help:

- When solving a specific issue, it's good to reference the issue, like this: https://github.com/ACM-UoA-Student-Chapter/dil/issues/7 (you basically just paste the link). In that way, one that reads the PR can immediately get what it is about. Also, the Github references the PR in the issue, so anyone reading the issue can see PRs trying to solve it.

- When a PR is separated into tasks, it's good to do separate PRs for each task. This is generally good because the smallest the PR -> the easier to review -> more possible that someone will take a look. Specifically in this project, task lists are good because more than one people can work on each issue. One might not be able to do the whole grammar but they might be able to just do the modules let's say.
So, if you please, take one task, e.g. expressions or whatever you like, and do a PR just on that. Then follow with the rest.

I have read your grammar, I will follow with comments on it in the separate PRs.

PR #14, opened 2019-11-29
 Jump started the Module part of the grammar

Doryanni:
#7 Added the module layout, structs, function definitions, and some basic types.

### DIFF
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -7,8 +7,9 @@ Declaration = StructDecl | FunctionDef;
 FunctionDef = "func", Identifier, "(",  FuncDefParamList, ")", "{",  { Statement }, "}";
 
 
-FuncDefParamList = [ParamIdentifierList, ":", Type, { ",", ParamIdentifierList, ":", Type }];
+FuncDefParamList = [ParamSet, { ",", ParamSet }];
 
+ParamSet = ParamIdentifierList, ":", Type;
 
 ParamIdentifierList = Identifier, { ",", Identifier };
 
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -1,25 +1,22 @@
-Module = { [FunctionDef], [StructDef] }, "EOF";
+Module = { Declaration }, ? EOI ?;
 
 
-FunctionDef = "func", Identifier, "(",  [FuncDefArgList], ")", "{",  { [Statement] }, "}";
+Declaration = StructDecl | FunctionDef;
 
 
-FuncDefArgList = IdentifierList, ":", Type, FuncDefArgListT;
-
-FuncDefArgListT = ",", IdentifierList, ":", Type, FuncDefArgListT
-                | ;
+FunctionDef = "func", Identifier, "(",  FuncDefParamList, ")", "{",  { Statement }, "}";
 
 
-IdentifierList = Identifier, IdentifierListT;
-
-IdentifierListT = ",", Identifier, IdentifierListT
-                | ;
+FuncDefParamList = [ParamIdentifierList, ":", Type, { ",", ParamIdentifierList, ":", Type }];
 
 
-StructDef = "struct", Identifier, "{", { [VarDeclStmt] }, "}";
+ParamIdentifierList = Identifier, { ",", Identifier };
 
 
-VarDeclStmt = Type, Identifier, ";";
+StructDecl = "struct", Identifier, "{", { StructField }, "}";
+
+
+StructField = Identifier, ":", Type, ";";
 
 
 Type = "int"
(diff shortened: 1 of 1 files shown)


baziotis:
Tnx for the PR. Here are some notes:

- An identifier is not a literal. It's a terminal having a specific possible structure (e.g. can start with underscore or letter, following underscore or letter or number etc. All that expressed in EBNF). But since most know what an identifier can be (same for integer literals), that can be skipped.

- The `VarDeclStmt` is used in structure declarations, so it's a bad name. :) Because it's neither
a variable declaration nor a statement. You could name this e.g. `AggregateField` or `StructField`.
Then second, and that's not your fault since I've given no examples about structs, the aggregate fields probably will not be declared as `int a` but `a: int`. So, first the name, then the type separated by `:`. That's not of high priority right now as it may not be exactly like that in the end.

- This `{ [something] }` is verbose. `{ }` denotes repetition, meaning 0 or more times repeated. So,
you could just have `{ something }`.

- The `IdentifierList` is _probably_ a bad name (probably because we might find use elsewhere).
In any case, it seems somewhat complicated to me. I think you can just do:
`ParamIdentifierList = Identifier { ",", Identifier };`.

- As above, I think `FuncDefArgList` could be:
`FuncDefArgList = { ParamIdentifierList ":" Type }`. Also, and that's nitpicking but it has some
importance, it would be better to name this: `FuncDefParamList` because arguments are what
is passed to the function while params is the formal declaration. For example:
```
func test(a, b : int) { ... }
...
test(1, 2);
```
1,2 is the argument list, while (a, int), (b, int) i.e. pairs of identifier and type is the parameter list (or
the formal paramater list). Other times the argument list is called the _actual_ parameter list.
I think it's more clear to say arguments and parameters.

- By using the rule for param list above you can just do `"(" FuncDefParamList ")"` because
we used repetition so we don't need the optional (because with repetition, you could choose to
have it 0 times). All that in `FunctionDef`.

- In `Module` `"EOF"` should be `EOF` or `<EOF>` denoting a special terminal,
the end of file (and actually, better name it EOI for end of input which is more general).
Otherwise, `Module` is a very clever rule but I think it might complicate things. By just looking
at it, I assumed that maybe `FunctionDef` and `StructDef` have some connection, but they don't.
I think it is clearer something like: A declaration can be a struct declaration or a function definition.
A module can have 0 or more declarations.

- `StructDef` would be better as `StructDecl`. Nitpicking again but a definition (as in the functions)
is different from a declaration.

Great job, I'm waiting for an update!

Doryanni:
> FuncDefArgList = { ParamIdentifierList ":" Type }

I believe that is incorrect due to a lack of a comma. Shouldn't it be: 

`FuncDefParamList = [ParamIdentifierList, ":", Type, { ",", ParamIdentifierList, ":", Type }];`

> In Module "EOF" should be EOF or <EOF> denoting a special terminal

I (perhaps incorrectly) assumed that we were following the ISO/IEC 14977 standard, so I think that the use of '??'  would be a more correct approach.

baziotis:
> > FuncDefArgList = { ParamIdentifierList ":" Type }
> 
> I believe that is incorrect due to a lack of a comma. Shouldn't it be:
> 
> `FuncDefParamList = [ParamIdentifierList, ":", Type, { ",", ParamIdentifierList, ":", Type }];`
> 

Yeah, thanks! It's weird how I forgot to add the repetition in the end. In my defense for the comma,
I've not been used to writing grammars that way. :)
My way is a little bit like BNF, with some regular expression additions. I prefer yours
since it is very standard. (but you know, I may have to do a quick rewrite just to be sure
I don't miss stuff like this).

> > In Module "EOF" should be EOF or  denoting a special terminal
> 
> I (perhaps incorrectly) assumed that we were following the ISO/IEC 14977 standard, so I think that the use of '??' would be a more correct approach.

I agree, `? EOF ?` is better.

EDIT: Actually, `? EOI ?`.

grammar.ebnf
```diff
@@ -0,0 +1,28 @@
+Module = { Declaration }, ? EOI ?;
+
+
+Declaration = StructDecl | FunctionDef;
+
+
+FunctionDef = "func", Identifier, "(",  FuncDefParamList, ")", "{",  { Statement }, "}";
+
+
+FuncDefParamList = [ParamIdentifierList, ":", Type, { ",", ParamIdentifierList, ":", Type }];
```
baziotis: You may want to have a rule for `ParamIndentifierList, ":", Type`, like `ParamSet` or smth. It's good either way, choose as you like.

baziotis:
@WilhelmStein Just saw the changes. Really good job! I hope it seems as beautiful to you as it seems to me. :)
Update me on the one comment I did (either you like the addition or not) and it will get merged.
I'll try to find some time to see the other PR today.

grammar.ebnf
Doryanni: Yeah, it is much more readable your way, to be honest, I'll change it promptly.

PR #15, opened 2019-11-29
Jump started the statement part of the grammar

Doryanni:
#7 Added block, while, assignment, array assignment and if statements.

### DIFF
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -1,21 +1,18 @@
-statement = "{", { [statement] }, "}"
-          | type, identifier, ";"
-          | "while", "(", expr, ")", "{", { [statement] }, "}"
-          | lVal, "=", expr, ";"
-          | lVal, "[", expr, "]", "=", expr, ";"
-          | "if", "(", expr, ")", statement;
+Statement = "{", { Statement }, "}"
+          | Type, Identifier, ";"
+          | "while", "(", Expression, ")", "{", { Statement }, "}"
+          | Lval, "=", Expression, ";"
+          | Lval, "[", Expression, "]", "=", Expression, ";"
+          | "if", "(", Expression, ")", Statement;
 
 
-lVal = identifier, lValT;
-
-lValT = ".", identifier, lValT
-      | ;
+Lval = Identifier, { ".", Identifier };
 
 
-type = "int"
+Type = "int"
      | "int", "[", "]" 
      | "bool"
-     | identifier;
+     | Identifier;
 
 
-identifier = ? identifier_literal ?;
+Identifier = ? IDENTIFIER ?;
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -28,7 +28,7 @@ Statement = "{", { Statement }, "}"
           | Identifier, "(", FuncCallArgList, ")", ";" 
           | "return", Expression, ";";
 
-FuncCallArgList = [Identifier, { ",", Identifier }];
+FuncCallArgList = [Expression, { ",", Expression }];
 
 
 Type = "int"
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -24,7 +24,11 @@ Statement = "{", { Statement }, "}"
           | "while", "(", Expression, ")", Statement
           | Identifier, "=", Expression, ";"
           | Identifier, "[", Expression, "]", "=", Expression, ";"
-          | "if", "(", Expression, ")", Statement;
+          | "if", "(", Expression, ")", Statement
+          | Identifier, "(", FuncCallArgList, ")", ";" 
+          | "return", Expression, ";";
+
+FuncCallArgList = [Identifier, { ",", Identifier }];
 
 
 Type = "int"
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -21,14 +21,11 @@ StructField = Identifier, ":", Type, ";";
 
 
 Statement = "{", { Statement }, "}"
-          | Type, Identifier, ";"
-          | "while", "(", Expression, ")", "{", { Statement }, "}"
-          | Lval, "=", Expression, ";"
-          | Lval, "[", Expression, "]", "=", Expression, ";"
+          | Type, Identifier, { ",", Identifier } ";"
+          | "while", "(", Expression, ")", { Statement }
+          | Identifier, "=", Expression, ";"
+          | Identifier, "[", Expression, "]", "=", Expression, ";"
           | "if", "(", Expression, ")", Statement;
-
-
-Lval = Identifier, { ".", Identifier };
 
 
 Type = "int"
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -26,7 +26,7 @@ Statement = "{", { Statement }, "}"
           | Identifier, "[", Expression, "]", "=", Expression, ";"
           | "if", "(", Expression, ")", Statement
           | Identifier, "(", FuncCallArgList, ")", ";" 
-          | "return", Expression, ";";
+          | "return", [Expression], ";";
 
 FuncCallArgList = [Expression, { ",", Expression }];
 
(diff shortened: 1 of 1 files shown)


Doryanni:
Making some changes currently as per the post in #14

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
 StructField = Identifier, ":", Type, ";";
 
 
+Statement = "{", { Statement }, "}"
+          | Type, Identifier, ";"
```
baziotis: What's this?

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
...
+          | "while", "(", Expression, ")", "{", { Statement }, "}"
+          | Lval, "=", Expression, ";"
+          | Lval, "[", Expression, "]", "=", Expression, ";"
+          | "if", "(", Expression, ")", Statement;
```
baziotis: Please make curly braces obligatory.

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
...
+
+
+Lval = Identifier, { ".", Identifier };
```
baziotis: This is a good thought but I propose that we restrict that to just `Identifier` in the start. Then we can start doing more complicated things.

baziotis:
Please write a PR description that makes clear what parts of the grammar this PR handles.

grammar.ebnf
Doryanni: It's a block statement. It opens up a new scope and can be used in conjunction with if and while statements. Although now I see that my code in while loops obligates braces whilst the if does not.

I'll also make it so that the variable definition works with multiple variables per type.

grammar.ebnf
Doryanni: Fair enough.

grammar.ebnf
Doryanni: Is it really necessary? One-liner ifs are quite common and putting obligatory braces seems a bit excessive and makes the code a bit harder to read, especially for people new to programming, who are this language's target audience (even if it's a mock project :p )

grammar.ebnf
Doryanni: I pushed a new commit making the if and while consistent. If you believe it's better to have mandatory braces, I'll add them.

baziotis: > It's a block statement.

I'm referring to line 24. It seems like a variable _declaration_. Which is... a declaration, not a statement :). We may allow "declare anywhere" but for the time being, we'll just have all variable declarations at the start.

grammar.ebnf
baziotis: Well, for me it makes the code harder to read to _not_ put obligatory braces (not in all cases, but in most in my experience).
However it's true that most people seem to like it so I agree, we'll keep it your way.

grammar.ebnf
Doryanni: Oh fair enough. You marked both lines so I assumed you meant both.

baziotis: Actually I didn't and as I said [here](https://github.com/ACM-UoA-Student-Chapter/dil/pull/16#discussion_r352233741) I don't know of a way to make Github mark only one specific line.

baziotis:
Please use consistently the final `;` in the rules.
Other than that, we need `return` statements and function calls and I think we're done.
The latter is subtle because a function call is also an expression. Languages like C / C++
couple statements and expressions so that an expression can be used as a statement. For example:
```
Statement = ...
          | Expression
```
This is not the best practice so for now, we can just make the rule for function calls and make
only that be able to be a statement.

Doryanni:
> Please use consistently the final `;` in the rules.

To what do you refer to?

baziotis:
> > Please use consistently the final `;` in the rules.
> 
> To what do you refer to?

Nothing, my bad. Me trying to get used to EBNF sometimes complicates things :)

Doryanni:
I tried changing my commit message, that is why multiple commits appeared.

spy-t:
@WilhelmStein take a look at [amend](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---amend)

baziotis:
Also in [squash](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/).
Experiment if you like, this is why we're making this mock project in the first place.

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
...
 
+Statement = "{", { Statement }, "}"
+          | "while", "(", Expression, ")", Statement
+          | Identifier, "=", Expression, ";"
+          | Identifier, "[", Expression, "]", "=", Expression, ";"
+          | "if", "(", Expression, ")", Statement
+          | Identifier, "(", FuncCallArgList, ")", ";" 
+          | "return", Expression, ";";
+
+FuncCallArgList = [Identifier, { ",", Identifier }];
```
baziotis: This is unfortunately not correct. Say you write this call: `a(1, 2)`.

Doryanni: Oops, silly mistake. Will refrain from posting so eagerly :p

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
...
 
+Statement = "{", { Statement }, "}"
+          | "while", "(", Expression, ")", Statement
+          | Identifier, "=", Expression, ";"
+          | Identifier, "[", Expression, "]", "=", Expression, ";"
+          | "if", "(", Expression, ")", Statement
+          | Identifier, "(", FuncCallArgList, ")", ";" 
+          | "return", Expression, ";";
+
+FuncCallArgList = [Expression, { ",", Expression }];
```
baziotis, line 33: Well done!

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
 StructField = Identifier, ":", Type, ";";
 
 
+Statement = "{", { Statement }, "}"
+          | "while", "(", Expression, ")", Statement
+          | Identifier, "=", Expression, ";"
+          | Identifier, "[", Expression, "]", "=", Expression, ";"
+          | "if", "(", Expression, ")", Statement
+          | Identifier, "(", FuncCallArgList, ")", ";"
```
baziotis, line 28: For educational purposes in case you don't have it in your mind, in languages like C / C++, the `Identifier` here should be an expression, because in `p->a(1, 2)`, `a` might be a function pointer. And a function pointer can result from arbitrary expressions. In C++ it can result from method referencing etc.
Here we don't have function pointers or methods (i.e. member functions) so no need for that.

grammar.ebnf
```diff
@@ -20,8 +20,19 @@ StructDecl = "struct", Identifier, "{", { StructField }, "}";
 StructField = Identifier, ":", Type, ";";
 
 
+Statement = "{", { Statement }, "}"
+          | "while", "(", Expression, ")", Statement
+          | Identifier, "=", Expression, ";"
+          | Identifier, "[", Expression, "]", "=", Expression, ";"
+          | "if", "(", Expression, ")", Statement
+          | Identifier, "(", FuncCallArgList, ")", ";" 
+          | "return", Expression, ";";
```
baziotis: Just make the `Expression` optional since we can have functions that return nothing.

grammar.ebnf
Doryanni: Yes, that is why I did not make it an expression in the first place. Thanks for the note nonetheless.

baziotis:
I added an important commit. It was a little subtle and I hadn't anticipated it so I thought better do it myself and add an explanation.
I think we're ok, thanks again. Merging!

baziotis:
There were some problems when this got merged with Git. I tried to solve it by squashing into one.

PR #16, opened 2019-11-29
a trivial implementation for string interning

J1635:
For start this is a trivial implementation for string interning. I will post a better one using a hash table.

### DIFF
--- a/samples/string_interning.cpp
+++ b/samples/string_interning.cpp
@@ -1,0 +1,18 @@
+#include <vector>
+#include <string>
+
+char* add_string(char *string, std::vector<char*> &old_strings) {
+    //check if the string is stored
+    for (size_t i = 0; i < old_strings.size(); i++) {
+        if (!strcmp(string, old_strings[i])) {
+            return old_strings[i];
+        }
+    }
+
+    //if it isn't stored create a new
+    char *new_string = new char[strlen(string) + 1];
+    strncpy(new_string, string, strlen(string) + 1);
+    old_strings.push_back(new_string);
+    
+    return new_string;
+}


baziotis:
Cool PR, I'll follow up with some feedback on the code.

samples/string_interning.cpp
```diff
@@ -0,0 +1,18 @@
+#include <vector>
+#include <string>
```
baziotis, line 2: Do we need that?

samples/string_interning.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+char* add_string(char *string, std::vector<char*> &old_strings) {
+    //check if the string is stored
```
baziotis: `add_string` is probably not the best name in the world :) It's quite generic, I think something like `intern_string` is better.

samples/string_interning.cpp
```diff
@@ -0,0 +1,18 @@
+#include <vector>
+#include <string>
+
+char* add_string(char *string, std::vector<char*> &old_strings) {
```
baziotis, line 4: - `add_string` is probably not the best name in the world :) It's quite generic, I think something like `intern_string` is better.
- You don't want the data structure that holds the saved strings to be visible to the caller. In this case, `old_strings`. The caller should pass a string and get a string.
- You can try returning and passing `const char *`. It's not that `const` is that useful in C / C++, but it just makes it somewhat clearer and safer that the strings are _only_ for reading.

samples/string_interning.cpp
```diff
@@ -0,0 +1,18 @@
...
+    //check if the string is stored
+    for (size_t i = 0; i < old_strings.size(); i++) {
+        if (!strcmp(string, old_strings[i])) {
```
baziotis, line 7: One good way to optimize this a little further is to save the length of each string apart from the string itself. That way, you can start comparing the strings only if their lenghts match. Because most of the time their lenghts won't match and so you'll just go to the next.

samples/string_interning.cpp
```diff
@@ -0,0 +1,18 @@
...
+            return old_strings[i];
+        }
+    }
+
+    //if it isn't stored create a new
+    char *new_string = new char[strlen(string) + 1];
+    strncpy(new_string, string, strlen(string) + 1);
```
baziotis, line 14: Cache the length. Probably the compiler will do it for you, but here it also makes the code more cryptic. The length is not supposed to change and so we don't have to call `strlen` 2 times.

baziotis:
I almost forgot. Add some code that tests this, in another function.

samples/string_interning.cpp
J1635: Because I'm using strlen, strncpy and vector.

baziotis: Oh yes, my bad for the `<string>`. Btw, I wanted to only mention line 2. But Github shows more than one lines and I don't know how to only show a single line. So, in this case it seems like I also referenced line 1 (i.e. `<vector>`) but it was not intended.
I was very confused on the start with all this.

J1635: Ok good to know

baziotis:
I missed another thing: When doing PRs, open a new branch for each new PR (tell if you don't know how to do that), don't do them in `master`. That is for your own convenience. Generally you might have to do force pushes and other things to make your master stay in sync with the original (i.e. the upstream) master. So, prefer a different branch. It's not necessary for now, but if you want to try it, you can open a new PR.

J1635:
Follow up: https://github.com/ACM-UoA-Student-Chapter/dil/pull/18

Issue #5, opened 2019-11-25
Generic hash table for data identified by a unique pointer


baziotis:
Write a generic [linear probing](http://www.cs.rmit.edu.au/online/blackboard/chapter/05/documents/contribute/chapter/05/linear-probing.html) hash table for data that are unique to a string identifier.

**Why linear probing?** 
Linear probing is fast! Period. It is optimized for cache locality as the data are contiguous.
What is more, the operations that are hazardous for a LP HT won't be needed for a compiler.
Those are deletion and non-constant length data. With the last one I mean that when
the hash table is create, we will know exactly how many data it will save.

**Structure**
The data that we will save will be searched by an unique pointer (specifically `a const char *`).
This is a topic of another issue but essentially all the identifiers will be string interned. That is,
equal strings point to the same memory. For example, let's say we save "foo" as a `const char *` in one place and in another. These 2 `const char *` pointers will be equal. What this means
is that essentially the hash table will treat those strings as integers (i.e. the values of pointers).

We will keep 2 parallel arrays. One that will save the actual data and another that saves the
pointers. In that way, we can do the searching in the pointers (very fast) and incur the penalty
only when we find the correct id. At this point we will return the parallel array pointer.

For this hash table we should constrain the data to only be pointers. That's important to save
memory in unused places in the data array because of linear probing (i.e. the hash table won't be full of course).

spy-t:
Sorry if this has been made clear before but because I'm intending to start with this issue, will the implementation language be C or C++?

baziotis:
C++, I haven't explicitly mentioned it anywhere. Best of luck with this issue, post any questions you might have.

**_Note on interface decisions_** (you might want to skip it):
Clarification: When I'm referring to "interface", I mean the way any abstracted piece of code interacts with the outer world. For example, a function is an abstraction: It does something but you don't necessarily know / care _how_. The interface of the function (and what you should definitely care about) is: what it does, what it gets as input, what it returns as output. In the same way, for a class are its methods and properties.

The best way to decide the correct interface is _by writing the usage code first_ (as Casey Muratori says). That is, if you want to make a function that does the X, then instead of starting from _writing_ the function, start from _using_ the function. For example, if at some point in my code I need a function that gets an `X` and returns a `Y`, I will write:
```
X x = ...;
Y y = my_function(x);
```
without knowing how `my_function()` will be implemented.
[...]

PR #17, opened 2019-11-29
Implement the linear probing hash table

spy-t:
**Issue**: https://github.com/ACM-UoA-Student-Chapter/dil/issues/5
## Overview
This is a proof of concept implementation of the linear probing hash table issued by  . The implementation is very brittle at the moment and a lot of thing look unoptimized and/or unsafe but it should be enough for a start . I also took the liberty of organizing the workspace and creating a generic Makefile with sane defaults.

## Design Choices
The table is implemented as a `struct` instead of a `class` because a `class` seemed overkill. It is generic over a type `T` and should work for simple types as demonstrated by the tests. The hash function is unoptimized and should be changed to something better ASAP.


spy-t:
Let me also add that it would be wise creating a contribution guide and defining some "rules" for contributing. For a example where to put tests, where to put files/headers, how to use the Makefile or any other build system we choose etc.

baziotis:
Great work!

> Let me also add that it would be wise creating a contribution guide and defining some "rules" for contributing. For a example where to put tests, where to put files/headers, how to use the Makefile or any other build system we choose etc.

That's true but for the time being, it is wasted time. Tests will be placed in a specific directory when the code will start to get shape.

I'll follow-up with notes on the code. Please remove .gitignore, makefile and .gitkeep. They're just noise for now.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+#include <stdlib.h>
+#include <stdint.h>
+#include <assert.h>
+
+template <typename T>
+struct lp_hash_table {
+private:
+    const char **keys;
+    T *values;
+    int buckets;
```
baziotis: Better name this `nbuckets` or something it signals that it keeps a count. Because in code, something like:
`ht->buckets` is not clear whether it refers to the array of buckets or to the number of buckets.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+        buckets = elems;
+
+        // TODO: use our own allocator when it is ready.
+        keys = (const char **)malloc(buckets * sizeof(const char *));
+        values = (T *)malloc(buckets * sizeof(T));
+
+        for (int i = 0; i <= buckets; ++i) {
+            // We need to initialize the pointer array because the insertion loop will check on it's indices
+            keys[i] = nullptr;
+        }
```
baziotis: `calloc()` them ;)

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+    };
+
+    /*
+     * @return - 0 means everything went ok, -1 means there was an error
+     */
+    int insert(const char *key, T value) {
```
baziotis: Better use `true` / `false`.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+
+        int index = this->hash(key);
+        int i = index;
```
baziotis: Why not `int i = this->hash(key);` (I know... nitpicking)

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+            if (i == index)
+                return -1; // We have traversed the whole table and didn't find the element
+        }
+        return -1;
+    };
+
+    /*
+     * We apparently don't need this
+     * int remove(const char *key);
+     */
```
baziotis: Right!

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
+#ifndef DIL_LP_HASH_TABLE_H
+#define DIL_LP_HASH_TABLE_H
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <assert.h>
+
+template <typename T>
```
baziotis: Constrain `T` to be only of type pointer. This is C++ wizardry and I'm not the biggest fan of it. But in this case it's useful.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+     */
+    lp_hash_table(int elems) {
+        assert(elems > 0);
+
+        // TODO: maybe find a way to optimize this
+        buckets = elems;
+
+        // TODO: use our own allocator when it is ready.
+        keys = (const char **)malloc(buckets * sizeof(const char *));
+        values = (T *)malloc(buckets * sizeof(T));
```
baziotis: If you want to save `N` elements into a hash table, then using space of `N` is a bad idea. That's because the most important thing in hash table is the good distribution. Good distribution means that when you will _probe_ you will hit very few elements (hence the O(1) amortized). Using space of `N` elements means that essentially this is linear search, i.e. O(n). You have to "waste" some space to achieve good distribution.

That's where the _load factor_ is introduced: A load factor in what percentage is the hash table full. For example, if we're using only 70 out of 100 items, then it's 70%. Usually 70% is the limit for a good load factor. Above that the hash table is very slow (and it needs rebalancing and allocating more elements).

Since we know the number of elements to save, the load factor is constant (still, a good distribution is not guarranteed though). So, we can just take space say `1.3 * elems`;

_Note_: There is a lot more theory (and math) involved with these things than the simplistic explanation I gave but I assumed I'll bore you to death.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+
+    /*
+     * @return - 0 means everything went ok, -1 means there was an error
+     */
+    int insert(const char *key, T value) {
+
+        int index = this->hash(key);
+        int i = index;
+
+        while(this->keys[i] != nullptr) {
```
baziotis: Notice that what you're doing here is essentially what find does. I think it pays off to think of a way to not duplicate this code.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+        this->keys[i] = key;
+        this->values[i] = value;
+        return 0;
+    };
+
+    /*
+     * @return - 0 everything went ok. -1 something went wrong
+     * @param value - the value of the element under `key` or nullptr if it does not exist
+     */
+    int find(const char *key, T *value) {
```
baziotis: `find` can just return the element. Since we constrain this hash table to only keep pointers, if didn't find it, we can just reuturn `NULL`. Otherwise, the found pointer. So, `T find(const char *key) { ... }`.

include/lp_hash_table.h
```diff
@@ -0,0 +1,100 @@
...
+        // clear value from anything it might hold.
+        int index = this->hash(key);
+
+        int i = index;
+
+        while(this->keys[i] != nullptr) {
+            if (this->keys[i] == key) {
+                // TODO: this seems unsafe
+                *value = this->values[i];
```
baziotis: You may want to throw an `assert` that it is not `NULL`. Generally, throw as many `assert`s as it is necessary, they're very important. Plus, we will overwrite the assert in optimized builds so no overhead.

baziotis:
Sorry, didn't have time to check the tests. I think that those comments can keep you going. I'll try to take a look tomorrow.

include/lp_hash_table.h
spy-t: Yeah I forgot we had this stuff in C++

include/lp_hash_table.h
spy-t: It's fine, I love big boring explanations

include/lp_hash_table.h
spy-t: I have this underway already but I wanted to push as fast as possible because of the workspace structure I chose. That way if anyone else wants to contribute code they could have a starting point.

include/lp_hash_table.h
spy-t: I thought about this, actually it was the original implementation, but I liked the consistency the API would have in it's signatures. Especially if we used enum error codes and stuff like `perror` like the standard library. Moreover I didn't constrain `T` to be a pointer (bear in mind that I haven't used C++ in 2 years) so I guess by doing that returning the pointer here would be a sane choice.

include/lp_hash_table.h
spy-t: Hmmm since a hash function is a pure function, meaning that `index == i == hash(key)`,  wouldn't calling it 2 times here be redundant?

baziotis: Yes, but what I meant is to not use `index` at all. Then I realized that `index` is used as a saved value to check for wrapping, so we need it. Scratch that. :)

include/lp_hash_table.h
baziotis: Of course, either way, it's quite good this way!

include/lp_hash_table.h
baziotis: > Moreover I didn't constrain T to be a pointer

Of course. This wasn't the case as you initially did it, so returning NULL was not an option.

> Especially if we used enum error codes and stuff like perror like the standard library

Oh no, please! :) The `perror` et al of error handling is really bad. They use a global variable for all that and it's just a pain to work with. For `enum` codes, that's interesting, but here basically there are 3 cases we care about:
1) The value was found, return it.
2) The value was not found, return `NULL`.
3) Something went wrong, `assert`, halt and catch fire. :) (because something should never go wrong).

spy-t: Fair enough. I'll do it that way

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+
+    /*
+     * @return - Copy of the value under key or NULL if it does not exist
+     */
+    T* find(const char *key) {
+
+        int i = find_index(key);
+        if(i != -1) {
+            if (this->keys[i] != nullptr) {
+                assert(this->keys[i] != nullptr);
```
baziotis: You have just tested above that `this->keys[i]` is not `nullptr`. No need to `assert` again. What it would be good to assert that is not nullptr is `this->values[i]`.

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+            this->values[i] = value;
+            return true;
+        }
+        return false;
+    };
+
+    /*
+     * @return - Copy of the value under key or NULL if it does not exist
+     */
+    T* find(const char *key) {
```
baziotis: This should return `T` because we're returning values, not pointers to values (it just happens that the values _are_ pointers).

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+
+        int i = find_index(key);
+        if(i != -1) {
+            if (this->keys[i] != nullptr) {
+                assert(this->keys[i] != nullptr);
+                return &this->values[i];
```
baziotis: Save as above, this should return `this->values[i]`.

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+            } else if (this->keys[i] == nullptr) {
```
baziotis: Just else. You have already tested if it is not nullptr. So, if we reached here, it means it is null.

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+     * Constructor. Assumes that `elems > 0`
+     */
+    lp_hash_table(int elems) {
+        assert(elems > 0);
+
+        this->nbuckets = elems * 1.3;
+
+        // TODO: use our own allocator when it is ready.
+        keys = (const char **)calloc(this->nbuckets, sizeof(const char *));
+        values = (T *)malloc(this->nbuckets * sizeof(T));
```
baziotis: You may `calloc` them as well since it's in your best interest to be null if you don't change them.

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+
+    };
+
+    /*
+     * @return. true - everything went ok. false - something went wrong
+     */
+    bool insert(const char *key, T value) {
+
+        int i = find_index(key);
+        if(i != -1 && this->keys[i] != key) {
```
baziotis: For the time being, you can `assert` that `i != -1` (and maybe make `insert` return `void`). Because we should always be able to find a position since the number of elements to be saved is constant.
~~The second check is not needed. :)~~ Scratch that, `assert` both of them.

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+        return false;
+    };
+
+    /*
+     * @return - Copy of the value under key or NULL if it does not exist
+     */
+    T* find(const char *key) {
+
+        int i = find_index(key);
+        if(i != -1) {
```
baziotis: Try to write this as:
```
if (i == -1) {
    return NULL;
}
... // rest of code
```
to reduce nesting.

baziotis:
Please remove .gitignore, .gitkeep and makefile.
Also, and I will write this when the coding guide is ready, pass this through clang-format (ask me if you don't know).

include/lp_hash_table.h
spy-t: That's what I was intending to write :(

include/lp_hash_table.h
```diff
@@ -0,0 +1,90 @@
...
+#include <stdint.h>
+#include <assert.h>
+
+
+template <typename T>
+struct lp_hash_table {
+private:
+    const char **keys;
+    T *values;
+    int nbuckets;
```
baziotis: Use `ssize_t` for this as the `int` can restrict us.

include/lp_hash_table.h
baziotis: Haha no worries :)

include/lp_hash_table.h
spy-t: Why not `size_t`? AFAIK `ssize_t` is just a signed `size_t` and `nbuckets` shouldn't be negative. Am I missing something?

baziotis: You're correct in that but you might not think the problems of unsigned integers. They have edge cases which a lot of people don't consider when writing code. A simple example:
```
for (size_t j = ...; j >= 0; --j)  // This results in an infinite loop
```

include/lp_hash_table.h
```diff
@@ -0,0 +1,86 @@
...
+
+template <typename T> struct lp_hash_table {
+  static_assert(std::is_pointer<T>::value, "T must be a pointer");
+
+private:
+  const char **keys;
+  T *values;
+  size_t nbuckets;
+
+  unsigned int hash(const char *key) {
```
baziotis: This should return `ssize_t` as it represents an index.

include/lp_hash_table.h
```diff
@@ -0,0 +1,86 @@
...
+    // TODO: implement a better hashing method
+    return ((uintptr_t)key % this->nbuckets);
+  }
+
+  int find_index(const char *key) {
```
baziotis: This should return `ssize_t` as well.

include/lp_hash_table.h
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    return i; // We found an empty spot.
+  }
+
+public:
+  /*
+   * Constructor. Assumes that `elems > 0`
+   */
+  lp_hash_table(int elems) {
```
baziotis: `ssize_t` in the elems.

include/lp_hash_table.h
```diff
@@ -0,0 +1,86 @@
+#ifndef DIL_LP_HASH_TABLE_H
+#define DIL_LP_HASH_TABLE_H
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <type_traits>
+
+template <typename T> struct lp_hash_table {
+  static_assert(std::is_pointer<T>::value, "T must be a pointer");
```
baziotis: Ok, this difficult to think about but check the [is_pointer doc](https://en.cppreference.com/w/cpp/types/is_pointer).
It says: Checks whether T is a pointer to object or _a pointer to function_. We don't want to handle the latter. It's not of great importance to check that but if you want to think about it, you may do.

tests/lp_hash_table_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+};
+
+bool test_hash_table() {
+  lp_hash_table<test_type *> table = lp_hash_table<test_type *>(1);
+
+  test_type *t = new test_type(1);
+  table.insert("key", t);
+
+  test_type *r;
+  r = table.find("key");
```
baziotis: That's not correct. Remember that we're hashing with _pointers_ (i.e. the keys are pointers). So, to hash and get the same value, we have to pass the same pointer (this is the whole point of this hash table :) ).
However, `"key"` in `insert` and `"key"` here are guaranteed to be in the same memory (i.e. same pointer).

Funny that if you ran this in optimized mode, then the compiler does string interning (which you might have seen around as another issue) on the string literals (meaning not while processing them, as we will do, but when it generates them).
That is, if 2 string literals in 2 (or more) separate locations are the same, it will save them only once and so here,
you will actually pass the same pointer.

baziotis: Same problem in the other tests :/ Which otherwise are good.

tests/main.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+int main() {
+  std::cout << "Hash Table" << std::endl;
+  bool res = test_hash_table();
+  assert(res == true);
+
+  res = test_many_key_insertion();
+  assert(res == true);
+
+  std::cout << "\x1b[0;32m Hash table tests passed! \x1b[0m" << std::endl;
```
baziotis: IIRC this colors the message. Funny but please don't do that because it's part of another issue and we don't want to spoil it. :P

spy-t: My bad sorry

tests/lp_hash_table_test.cpp
spy-t: In other words this test should fail if the implementation was indeed correct right? Except if it was running in optimized mode of course.

baziotis: Well, except if the compiler decided to do string interning in unoptimized mode. I don't remember the last time I checked this, I'll come back. Otherwise yes, it should fail but I haven't run this.

spy-t: Ok the way I'm hashing seems right, I'm using the pointer so the 2 "key"'s should be different unless of course interning was used. I'm compiling with -Wall and -g if that helps

baziotis: Well, last time I checked GCC did not do string interning. But now it seems it does, at least for trivial cases. One test you can run yourself:
```
#include <stdio.h>
void test(const char *s) {
  printf("%p\n", s);
  ++s;
}
int main(void) {
  test("key");
  printf("whatever message\n");
  test("key");
}
```
So, even with `-g`, this will intern `"key"` and that's why your code works.

tests/main.cpp
baziotis: No problem.

tests/lp_hash_table_test.cpp
spy-t: Ok good that means that the hash function is working properly and that I just need to change the tests so they pass with interning turned off.

include/lp_hash_table.h
spy-t: ~~So `!std::is_pointer<std::function<T> *>::value` should work right?~~ Nope.

baziotis: Actually sorry for directing you in this path and making you waste your time. This is C++ magic that we should not be focusing on.
One workaround is this:
```cpp
#include <functional>

template <typename T> struct A {
  static const bool is_pointer = std::is_pointer<T>::value;
  static const bool is_function_pointer =
      is_pointer
          ? std::is_function<typename std::remove_pointer<T>::type>::value
          : false;
  static_assert(is_pointer && !is_function_pointer,
                "Template parameter should be of pointer type (but not a "
                "function pointer).");
};

A<int *> a;  // Pass
A<void (*)()> b;  // Fail
```
You may ignore that.

spy-t: It's ok the missing piece was `std::remove_pointer`.

include/lp_hash_table.h
```diff
@@ -0,0 +1,93 @@
+#ifndef DIL_LP_HASH_TABLE_H
+#define DIL_LP_HASH_TABLE_H
+
+#include <assert.h>
+#include <functional>
+#include <stdint.h>
+#include <stdlib.h>
+#include <type_traits>
+
+template <typename T> struct lp_hash_table {
```
baziotis: People choose a way to make clear that something is a struct by the naming.
That's usually either by putting a `_t` or using some different naming convention.
I will upload a coding guide soon and I propose that we use `snake_case` generally, but for structs `PascalCase`.
It seems to be the most effective convention I have worked with. So you may rename that to e.g. `IdHashTable` or something like that. It will probably change anyway when we find uses for it.

tests/lp_hash_table_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+  assert(r->val = 2);
+
+  r = table.find(k3);
+  assert(r->val = 3);
+
+  r = table.find(k4);
+  assert(r->val = 4);
+
+  r = table.find(k5);
+  assert(r->val = 5);
```
baziotis, line 69: Please check that do not exist and keys that have the same contents but are in different pointers.
We should also find a way to check that some asserts fire but that requires some more complex testing capabilities. So, we're leaving it for now.

include/lp_hash_table.h
```diff
@@ -0,0 +1,93 @@
+#ifndef DIL_LP_HASH_TABLE_H
+#define DIL_LP_HASH_TABLE_H
+
+#include <assert.h>
+#include <functional>
+#include <stdint.h>
+#include <stdlib.h>
+#include <type_traits>
+
+template <typename T> struct LpHashTable {
```
baziotis, line 32: Add some doc above about what this hash table does and how.

tests/lp_hash_table_test.cpp
spy-t: Can you rephrase that? I don't get it.

include/lp_hash_table.h
```diff
@@ -0,0 +1,93 @@
...
+   * Constructor. Assumes that `elems > 0`
+   */
+  LpHashTable(ssize_t elems) {
+    assert(elems > 0);
+
+    this->nbuckets = elems * 1.3;
+
+    // TODO: use our own allocator when it is ready.
+    keys = (const char **)calloc(this->nbuckets, sizeof(const char *));
+    values = (T *)calloc(this->nbuckets, sizeof(T));
```
baziotis, line 95: `assert` that we don't get `NULL` in both.

tests/lp_hash_table_test.cpp
baziotis: Oh yeah, I meant to say "Check keys that do not exist (in the hash table)". And also keys that have the same contents (i.e. `strcmp() will give 0`) but are in different memory locations (different pointers).

tests/lp_hash_table_test.cpp
```diff
@@ -64,3 +65,36 @@ bool test_many_key_insertion() {
 
   return true;
 }
+
+/*
+ * NOTE: this test will fail if the compiler implementation interns strings.
+ * IE if two `const char *` have the same address in memory this test will fail.
+ */
```
baziotis: Because of that, you allocate the memory for the strings dynamically. :)

spy-t: Fair point. I could do something like this
```c
  const char *k1 = "key1";
  char *k2 = (char *)malloc(5 * sizeof(char));
  std::memcpy(k2, k1, sizeof(k2));
```
but I think that this defeats the purpose of this test because it is a workaround for an implementation specific "problem". Shoud I remove this case all together or conditionally run it if the two pointers don't point at the same location?

include/lp_hash_table.h
```diff
@@ -7,6 +7,14 @@
...
 #include <type_traits>
 
+/*
+ * LpHashTable is a struct implementing a `linear probing hash table`.
+ * You can find more information about about that here: https://en.wikipedia.org/wiki/Linear_probing.
+ * It functions like a normal hash table but it handles collisions by linearly
+ * iterating over the hash table until an empty spot is found. The search is performed
+ * in a similar fashion. Because of the use case we are covering in the DIL compiler
+ * there is no need for a `delete` method and the size of the table is constant.
+ */
```
baziotis: Let me help in that a little. I think a good way to provide doc is to first of all split in the usage info and the internal / implementation info. That is, if you open the file, you should immediately be able to answer the questions: "What is this?", "How do I use it?", "How fast it is?". So, we may start as:
```
This is a map of const char * to pointer values (of any type) supporting insertion and search.
```
Notice that I didn't refer to a "hash table" because the fact that this is implemented as a hash table is not important to the user. Continuing...
```
Important details:
-- The number of elements to be saved is constant and is provided in its construction.
-- Keys: The contents of the memory pointed by the const char * passed are _not_ considered.
         Only its (pointer) value. That is, these pointers are handled as integer keys.
-- Values: Should be pointers to objects of any type.
Performance:
Insertion: O(1) amortized.
Search: O(1) amortized.
```
If you want to improve that, please do! Then, we'll continue to the implementation details.

spy-t: I misinterpreted what you meant by documentation. I thought that because it was source code documentation it should document the implementation rather than the higher level abstraction. I will work on it.

tests/lp_hash_table_test.cpp
baziotis: > but I think that this defeats the purpose of this test because it is a workaround for an implementation specific "problem".

Oh no, it doesn't! :) As you said, this is an implementation detail. The tests should not be based on things that are implementation-dependent! We should try our best against things like that.

baziotis: > ```c
>   const char *k1 = "key1";
>   char *k2 = (char *)malloc(5 * sizeof(char));
>   std::memcpy(k2, k1, sizeof(k2));
> ```

In this, you may want to try `strdup()`. This is wrong btw because the `sizeof(k2)` in
`memcpy()` should be 5. `sizeof(k2)` is the size of the _pointer type_, not the memory it points to. That is always 4 or 8 depending on the target.
Last, it's better to do an `strlen` rather than `5`.

include/lp_hash_table.h
baziotis: Both are part of the documentation. :)
Sure, you may put them in different places depending on the code though.

tests/lp_hash_table_test.cpp
spy-t: The point was the dynamic allocation yes. So what I should do is test both implementation cases right?

baziotis: The point is to have different memory locations. So:
> So what I should do is test both implementation cases right?

No. That is because let's say you want to test X. You don't try to test for all the different
implementation cases. You're just trying to know about them, so that you try to not be dependent in them. For example, in this case, you want to test that different pointers with the same contents are saved in different places. So, you just try to find a way to _always_ have different memory with the same contents, no matter the implementation. That is achieved e.g. by allocating dynamically.

Edit: Update me if that made sense or not.

spy-t: Yeah it was clear from the start I just rushed it at this point I guess.

baziotis: Good :)

include/lp_hash_table.h
```diff
@@ -0,0 +1,113 @@
...
+#define DIL_LP_HASH_TABLE_H
+
+#include <assert.h>
+#include <functional>
+#include <stdint.h>
+#include <stdlib.h>
+#include <type_traits>
+
+/*
+ * LpHashTable is a map of key-value pairs where keys are constant strings
```
baziotis: Ok, that's much better! :)
Please use "keys are `const char *`" just to not imply that we save strings in any way.
(I know you mention it below and that's good, but also people when trying to write code fast may not look on the notes initially and so one sentence can confuse them).

include/lp_hash_table.h
```diff
@@ -0,0 +1,113 @@
...
+
+    assert(keys != NULL);
+    assert(values != NULL);
+  };
+
+  /*
+   * @return. nothing. we are sure that we can insert
+   */
+  void insert(const char *key, T value) {
+
```
baziotis, line 107: Throw an `assert` here that `key` and `value` are not `NULL`.

include/lp_hash_table.h
```diff
@@ -0,0 +1,113 @@
...
+    assert(this->keys[i] != key);
+    this->keys[i] = key;
+    this->values[i] = value;
+  };
+
+  /*
+   * @return - Copy of the value under key or NULL if it does not exist
+   */
+  T find(const char *key) {
+
```
baziotis, line 120: `assert` here for the `key` that it is not `NULL`.

include/lp_hash_table.h
spy-t: It's funny because that's how I originally wrote it down but then I thought that it might show too much internal implementation haha.

baziotis: Well.. The user has to know that it saves `const char *` 'cause they have to pass one. :P 
As a general thought, it's kind of "artistic" how much info you give the user.
The convention is: "Abstract as much as you can from the user". That gets worse with all the OOP, getters, setters and the like.
I'm not a fan of it. I don't think the user should think they use some kind of magic.
Knowing some info about what the thing does means you can e.g. `reserve()` in a `std::vector` and save a lot of time.

spy-t: I agree

baziotis: And btw I forgot: You may now write the doc for the internals :)

include/lp_hash_table.h
```diff
@@ -84,6 +84,7 @@ template <typename T> struct LpHashTable {
    * @return. nothing. we are sure that we can insert
    */
   void insert(const char *key, T value) {
+    assert(key != NULL && value != NULL);
```
baziotis: Do it in separate asserts. I know this seems like nitpicking but you'll it very useful when one assert fires, to know immediately which thing was the problem.

baziotis:
Referring to your last commit:
`assert` is **_not_** compile-time assertion. It's a runtime assertion that fires when you run the program, not when you compile it.
`static_assert` is compile-time assertion.

baziotis:
Unless I'm missing something, after the doc for the internals is complete, we can merge that. Of course, when we're about to use it we'll come back to improve the hash function but it can wait.
Do me a favor and add a license and copyright notice with this template at the top of the file:
```cpp
/*
* Authors:   <YOUR_NAME>
*/
```

include/lp_hash_table.h
```diff
@@ -25,6 +30,12 @@
  * Search: O(1) amortized.
  */
 template <typename T> struct LpHashTable {
+  /*
+   * The map is implemented as a linear probing hash table. Keys are hashed by their
+   * pointer and are inserted into the table at the first empty spot (found linearly)
+   * in case the spot is taken. Find is implemented in a similar fashion. Delete is
+   * not implemented as we have no need for it.
+   */
```
baziotis: Please rephrase that a little. It's generally good, but I propose 2 small changes. First, explain the fact that we have 2 parallel arrays, one for the keys, one for the values. Second, disambiguate what happens in the case the spot is taken. So, maybe something like:
```cpp
/*
 * The map is implemented as a linear probing hash table. There are 2 parallel
 * arrays, one for keys and one for values. The keys are hashed by their pointer
 * and are inserted into the table at the first empty spot. If the hashed spot
 * is empty, we insert the key there and the value in its parallel spot.
 * Otherwise, we search linearly for the next empty key spot (i.e. linear probe)
 * and do the same thing. Find is implemented in a similar fashion. Delete is
 * not implemented as we have no need for it. Notes:
 *  - The load factor is a constant 70% (the number of elements in the table is
 * constant).
 *  - The values must be pointers. This constraint is imposed in order to save
 * space, as some of it is wasted to keep a load factor of 70%.
 */
```
Feel free to change that further.

spy-t: I'm going to copy-paste that so we can close the issue and merge because I'm not going to have enough time in the coming weeks. We can improve on that later when we actually need the map.

baziotis:
Ok, that's ready to be merged. Kudos mate, that was the hardest issue till now. :)

Issue #6, opened 2019-11-25
String interning


baziotis:
Interning strings means that we keep only one copy in memory of each unique string.
For example, let's say we want to save the string "foo". Then, we allocate some memory to save its chars. If we then want again to save "foo", instead of allocating new memory,
we point again to the same previous memory.
That means a couple of things:
- We don't waste memory.
- Remember that a string is essentially a pointer to char. If we keep only one copy around, then
pointers of the same string point to same memory. That means that a string is now identified
by an integer value (the value of the pointer). This makes the rest of the compiler _very_ fast
as it works with integers instead of whole strings.

For now it's better to start with the easiest way that you can think of implementing this.
Then, we'll improve it.

J1635:
I'd like to work on this one, can I get assignment?

J1635:
Are there any restrictions regarding what we can use to develop our programs. For example in cpp, can we use vector and String?

baziotis:
In general no, we just want sane C++. Now because this is an educational project and because vector and string will create more problems than they'll solve, please refrain from using them. I think that without them it will be easier.

**EDIT**: Sorry, I just realized that for an easy solution, the vector will make your life easier. You may use it.

baziotis:
Implemented in: https://github.com/ACM-UoA-Student-Chapter/dil/pull/18.

PR #18, opened 2019-11-30
String interning

J1635:
Issue: https://github.com/ACM-UoA-Student-Chapter/dil/issues/6
Continuation of PR: https://github.com/ACM-UoA-Student-Chapter/dil/pull/16
Added test. Edited code according to the comments.


baziotis:
You dont write "new pull request of...", you just say continuation of and paste the link of the previous ;).
You should not put your code in samples. Samples contain code written on DIL (sample code), _not_ of the compiler.
Also, add more testing. Otherwise, it looks good.

J1635:
Should I create a compiler directory to put my code inside?

baziotis:
Yes, good idea

compiler/test_intern_string.cpp
```diff
@@ -0,0 +1,16 @@
...
+  const char *input_1 = "test";
+  const char *input_2 = "test 2";
+
+  const char *address_1 = intern_string(input_1);
+  const char *address_2 = intern_string(input_2);
+
+  assert(address_1 == intern_string(input_1));
+  assert(address_2 == intern_string(input_2));
+  assert(address_1 != intern_string(input_2));
+  assert(address_2 != intern_string(input_1));
```
baziotis: The last 2 are not much useful but the other are good! There's not much to test in here because of most of the functionality is being handled by already tested code (e.g. `strlen`, `vector` etc.). But what other edge cases can you think of?

compiler/intern_string.cpp
```diff
@@ -0,0 +1,28 @@
+#include <string>
+#include <vector>
+
+typedef struct InternedStr {
```
baziotis: No need to `typedef` in C++.

baziotis:
> You dont write "new pull request of...", you just say continuation of and paste the link of the previous ;).

Ok that was probably misleading. I meant that you write that in the PR _description_, not in the title.
The title should be just what it does, e.g. Simple string interning

baziotis:
Great and as a final step you can close the other PR.

compiler/intern_string.cpp
```diff
@@ -0,0 +1,28 @@
+#include <string>
+#include <vector>
+
```
baziotis, line 15: Add doc about what this module will do.

compiler/test_intern_string.cpp
```diff
@@ -0,0 +1,17 @@
...
+  const char *input_2 = "test 2";
+
+  const char *address_1 = intern_string(input_1);
+  const char *address_2 = intern_string(input_2);
+  
+  assert(address_1 != NULL);
+  assert(address_2 != NULL);
+
+  assert(address_1 == intern_string(input_1));
+  assert(address_2 == intern_string(input_2));
```
baziotis, line 21: As I had commented somewhere else, what other edge cases can you think of?

compiler/intern_string.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <assert.h>
+
+/*
+* intern_string is a function which helps keeping only one copy in memory of each 
+* specific string.
+* intern_string parametre is the string we want to store
+* intern_string return value is a pointer
+* if we have already used the specific string the function returns the old pointer
+* if its a new string we allocate memory and return the new pointer
+*/
```
baziotis: Please use a better format. Something like:
```cpp
/*
 * This helps keep only one copy of every string memory. We pass it a string to
 * be saved. If the string is already saved, we return the old pointer.
 * Otherwise, we save it and return the new pointer. This is called interning a
 * string.
 * Params:
 *    string = The string to be interned.
 * Returns:
 *    A read-only pointer to a copy of `string`.
 */
```

compiler/test_intern_string.cpp
```diff
@@ -0,0 +1,25 @@
+#include "intern_string.h"
+#include <assert.h>
+
+int main(int argc, char const *argv[]) {
+  const char *input_1 = "test";
+  const char *input_2 = "test 2";
+  const char *input_3 = "";
+  const char *input_4;
```
baziotis: As it seems from the tests, you assume that this is initialized to `NULL` but this is not necessarily the case (that is because it has undefined value - it just _happens_ that most compilers (especially in unoptimized modes) initialize it to 0 for better predictability).
You can initialize it explicitly (i.e. ` = NULL;`).

baziotis:
Don't forget to resolve conversations when you push commit that addresses a review comment. It makes the life of the reviewer much easier ;)

baziotis:
Sorry, I forgot to prompt you to add a license / copyright notice. I added it for you, please verify that it is correct.

J1635:
Yes it's correct.

PR #21, opened 2019-12-02
First draft of expressions

Doryanni:
This is the first draft of expressions. I require some clarification with regard to arrays.
Specifically, do we only have the allocation expression `new int [ Expression ]` or do we plan on having arrays of different types, in the end? If it's the latter I have some amendments to do.

P.S. I might have gone a bit overboard with the expressions to make sure that when programming in DIL, it will take less code and time.

### DIFF
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -1,3 +1,10 @@
+(*
+* Copyright: Copyright Ioannis Cheilaris 2019.
+* License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+* Authors:   Ioannis Cheilaris, Stefanos Baziotis
+*)
+
+
 Module = {Declaration}, ? EOI ? ;
 
 Declaration = StructDecl | FunctionDef;
@@ -25,7 +32,7 @@ Statement = "{", {Statement}, "}"
           | "return", [Expression], ";";
 
 
-/*
+(*
   Note that we only need to support Identifier.Identifier and not an arbitrary
   number of .field, like a.b.c = 2 or a    .b.c.d = 2; That is because structs
   and arrays are always references, i.e. pointers. So, the above can be achieved
@@ -35,36 +42,47 @@ Statement = "{", {Statement}, "}"
   because we can achieve that as:
   Arr arr = a.b;  // arr is a pointer under the hood - any change to its
   elements will be reflected to a.b arr = a.b;
-*/
+*)
+
 LValue = Identifier, [ ".", Identifier ];
 
-Expression = ExpressionTerm 
-           | ParenthesisExpression
-           | FunctionCall
+Expression = FunctionCall
            | NotExpression
            | ArrayLookupExpression
-           | BinaryExpression
-
-BinOp = "+" | "-" | "*" | "/" | "%" | "&&" | "||" | "<" | "=<"| ">" | ">="| "=="| "!=";
-
-ParenthesisExpression = "(", Expression, ")";
+           | BinaryExpression;
 
 FunctionCall = Identifier, "(", FuncCallArgList, ")";
 
 FuncCallArgList = [ Expression, {",", Expression} ];
 
+(*
+  Expression precedence is simulated by using levels.
+  The higher the level, the bigger the precedence.
+*)
 
-BinaryExpression = ExpressionTerm, BinOp, ExpressionTerm;
+BinaryExpression = Level1Expr, [ "&&", Level1Expr, [BinaryExpression]];
 
-NotExpression = "!", Expression;
+Level1Expr = Level2Expr, [ "||", Level2Expr, [Level1Expr]];
 
-ArrayLookupExpression = ExpressionTerm, "[", ExpressionTerm, "]";
+Level2Expr = Level3Expr, [ ( "<" || "<=" || ">" || ">=" || "==" || "!="), Level3Expr, [Level2Expr] ];
+
+Level3Expr = Level4Expr, [ ("+" || "-"), Level4Expr, [Level3Expr]];
+
+Level4Expr = ExpressionTerm, [ ("*" || "/" || "%"), ExpressionTerm, [Level4Expr]];
+
+
+NotExpression = { "!" }, ExpressionTerm;
+
+ArrayLookupExpression = ExpressionTerm, "[", Expression, "]";
+
+ParenExpression = "(", Expression, ")";
 
 ExpressionTerm = Integer
-               | Identifier 
+               | LValue 
                | "true" 
                | "false" 
-               | "new", "int", "[", Expression,"]";
+               | "new", "int", "[", Expression,"]"
+               | ParenExpr;
 
 
 Type = "int"
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -74,5 +74,5 @@ Type = "int"
      | Identifier;
 
 
-Integer = ? INTEGER ?;
+Integer = ? INTEGER_LITERAL ?;
 Identifier = ? IDENTIFIER ? ;
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -25,10 +25,6 @@ Statement = "{", {Statement}, "}"
           | "return", [Expression], ";";
 
 
-FunctionCall = Identifier, "(", FuncCallArgList, ")";
-
-FuncCallArgList = [ Expression, {",", Expression} ];
-
 /*
   Note that we only need to support Identifier.Identifier and not an arbitrary
   number of .field, like a.b.c = 2 or a    .b.c.d = 2; That is because structs
@@ -43,14 +39,26 @@ FuncCallArgList = [ Expression, {",", Expression} ];
 LValue = Identifier, [ ".", Identifier ];
 
 Expression = ExpressionTerm 
-           | "(", Expression, ")"
+           | ParenthesisExpression
            | FunctionCall
-           | "!", Expression
-           | ExpressionTerm, "[", ExpressionTerm, "]"
-           | ExpressionTerm, InfixBinOp, ExpressionTerm;
+           | NotExpression
+           | ArrayLookupExpression
+           | BinaryExpression
 
-InfixBinOp = "+" | "-" | "*" | "/" | "%" | "&&" | "||" | "<" | "=<"| ">" | ">="| "=="| "!=";
+BinOp = "+" | "-" | "*" | "/" | "%" | "&&" | "||" | "<" | "=<"| ">" | ">="| "=="| "!=";
 
+ParenthesisExpression = "(", Expression, ")";
+
+FunctionCall = Identifier, "(", FuncCallArgList, ")";
+
+FuncCallArgList = [ Expression, {",", Expression} ];
+
+
+BinaryExpression = ExpressionTerm, BinOp, ExpressionTerm;
+
+NotExpression = "!", Expression;
+
+ArrayLookupExpression = ExpressionTerm, "[", ExpressionTerm, "]";
 
 ExpressionTerm = Integer
                | Identifier 
--- a/grammar.ebnf
+++ b/grammar.ebnf
@@ -2,7 +2,7 @@ Module = {Declaration}, ? EOI ? ;
 
 Declaration = StructDecl | FunctionDef;
 
-FunctionDef = "func", Identifier, "(", FuncDefParamList, ")", "{", {Statement},
+FunctionDef = "func", Identifier, "(", FuncDefParamList, ")", [":", Type] "{", {Statement},
     "}";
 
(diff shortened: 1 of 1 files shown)


baziotis:
> This is the first draft of expressions. I require some clarification with regard to arrays.
> Specifically, do we only have the allocation expression `new int [ Expression ]` or do we plan on having arrays of different types, in the end? If it's the latter I have some amendments to do.

We will, but let's start with just `int` arrays.

Mate my bad, it seems I have broken the formatting in the grammar file. Let me correct that
and you can rebase.

Doryanni:
Oh, I thought it was some sort of preference and thus I wrote the expressions in that style. Oh well, time to reformat a bit.

baziotis:
> Oh, I thought it was some sort of preference and thus I wrote the expressions in that style. Oh well, time to reformat a bit.

Sorry for that.

grammar.ebnf
```diff
@@ -2,7 +2,7 @@ Module = {Declaration}, ? EOI ? ;
 
 Declaration = StructDecl | FunctionDef;
 
-FunctionDef = "func", Identifier, "(", FuncDefParamList, ")", "{", {Statement},
+FunctionDef = "func", Identifier, "(", FuncDefParamList, ")", [":", Type] "{", {Statement},
```
baziotis: Good catch, I had totally forgot the return type.

grammar.ebnf
```diff
@@ -36,11 +42,37 @@ Statement = "{", {Statement}, "}"
...
+               | "new", "int", "[", Expression,"]";
+
 
 Type = "int"
      | "int", "[", "]"
      | "bool"
      | Identifier;
 
+
+Integer = ? INTEGER ?;
```
baziotis: `INTEGER_LITERAL` ;)

Doryanni: :)

grammar.ebnf
```diff
@@ -36,11 +42,37 @@ Statement = "{", {Statement}, "}"
...
+           | ExpressionTerm, "%", ExpressionTerm
+           | ExpressionTerm, "[", ExpressionTerm, "]"
+           | ExpressionTerm, "&&", ExpressionTerm
+           | ExpressionTerm, "||", ExpressionTerm
+           | ExpressionTerm, "<", ExpressionTerm
+           | ExpressionTerm, ">", ExpressionTerm
+           | ExpressionTerm, "=<", ExpressionTerm
+           | ExpressionTerm, ">=", ExpressionTerm
+           | ExpressionTerm, "==", ExpressionTerm
+           | ExpressionTerm, "!=", ExpressionTerm;
```
baziotis: You can compress all of those with something like:
```
BinOp = "+" | "-" | ...
BinaryExpression = ExpressionTerm, BinOp, ExpressionTerm
```
Then add the `BinaryExpression` as an `Expression`

WilhelmStein: Fair enough.

grammar.ebnf
```diff
@@ -36,11 +38,40 @@ Statement = "{", {Statement}, "}"
...
+Expression = ExpressionTerm 
+           | ParenthesisExpression
+           | FunctionCall
+           | NotExpression
+           | ArrayLookupExpression
+           | BinaryExpression
+
+BinOp = "+" | "-" | "*" | "/" | "%" | "&&" | "||" | "<" | "=<"| ">" | ">="| "=="| "!=";
+
+ParenthesisExpression = "(", Expression, ")";
```
baziotis: Better Parenthesized but you may want to write as Paren because it gets too verbose. Pick what you like.

grammar.ebnf
```diff
@@ -36,11 +38,40 @@ Statement = "{", {Statement}, "}"
...
 
+
+BinaryExpression = ExpressionTerm, BinOp, ExpressionTerm;
+
+NotExpression = "!", Expression;
+
+ArrayLookupExpression = ExpressionTerm, "[", ExpressionTerm, "]";
+
+ExpressionTerm = Integer
+               | Identifier
```
baziotis: We should let `LValue` here.

grammar.ebnf
```diff
@@ -36,11 +38,40 @@ Statement = "{", {Statement}, "}"
 */
 LValue = Identifier, [ ".", Identifier ];
 
+Expression = ExpressionTerm 
+           | ParenthesisExpression
+           | FunctionCall
+           | NotExpression
+           | ArrayLookupExpression
+           | BinaryExpression
```
baziotis: Ok, that change is good. But this does not consider precedence.

grammar.ebnf
Doryanni: You mean, add it to the ExpressionTerm rule?

baziotis: No need to add it. You can replace it with `Identifier` because an `LValue` can be an `Identifier`.

grammar.ebnf
Doryanni: Well, this merits further conversation. Should we opt for the following order of precedence?
```
# High Precedence
|    * \ %
|    + - 
|  <   =<    >=    ==    !=
|    ||    && 
# Low Precedence
```
Also, wouldn't precedence levels invalidate the following rule you proposed, or did you suggest it as a stepping stone?
```
BinaryExpression = ExpressionTerm, BinOp, ExpressionTerm;
```

baziotis: > Well, this merits further conversation. Should we opt for the following order of precedence?

Sorry if that was overwhelming. Yeah, it seems good. Just make `||` have higher precedence than `&&`.

> Also, wouldn't precedence levels invalidate the following rule you proposed, or did you suggest it as a stepping stone?

Yes, as a stepping stone because the thought is similar when precedence is added.

Doryanni: > Sorry if that was overwhelming.

Oh, worry not, I just like avoiding needless commits and changes ;)

baziotis: Okok :) Please do me a favor and add license / copyright notice in this file, with this template:
```cpp
/*
* Authors:   <YOUR_NAME>
*/
```
You may open a separate PR for that or leave it to the next grammar PR.

Doryanni: Isn't the copyright <YOUR_NAME> tag supposed to be for the project and not for the person writing it?

baziotis: Not in the way the Boost license has been written. You can have multiple copyright claims in different files. That's actually one of the reasons I chose BSL.

Doryanni:
Currently, expressions such as `Identifier [ true ]` or `new int [ true && false ]` are possible.
I am of the opinion that rectifying this might make the grammar more unreadable since we could easily have this pop up as an error during the semantic checking phase. What is your opinion on this?

grammar.ebnf
```diff
@@ -33,14 +42,54 @@ Statement = "{", {Statement}, "}"
...
   elements will be reflected to a.b arr = a.b;
-*/
+*)
+
 LValue = Identifier, [ ".", Identifier ];
 
+Expression = FunctionCall
+           | NotExpression
+           | ArrayLookupExpression
+           | BinaryExpression;
```
baziotis: Those should have a precedence as well. For example, the `[]` should be of higher precedence than `+`.

grammar.ebnf
```diff
@@ -33,14 +42,54 @@ Statement = "{", {Statement}, "}"
...
+FunctionCall = Identifier, "(", FuncCallArgList, ")";
+
 FuncCallArgList = [ Expression, {",", Expression} ];
 
+(*
+  Expression precedence is simulated by using levels.
+  The higher the level, the bigger the precedence.
+*)
+
+BinaryExpression = Level1Expr, [ "&&", Level1Expr, [BinaryExpression]];
```
baziotis: You don't need the `[BinaryExpression]`.  You can just write `BinaryExpression = Level1Expr, [ "&&", Level1Expr]`.
Also, `&&` should have higher precedence than `||`.
Lastly, you can name those in a more descriptive way, like `Level1Expr -> OrExpr`, `Level2Expr -> CmpExpr` etc.

grammar.ebnf
```diff
@@ -33,14 +42,54 @@ Statement = "{", {Statement}, "}"
...
+Level2Expr = Level3Expr, [ ( "<" || "<=" || ">" || ">=" || "==" || "!="), Level3Expr, [Level2Expr] ];
+
+Level3Expr = Level4Expr, [ ("+" || "-"), Level4Expr, [Level3Expr]];
+
+Level4Expr = ExpressionTerm, [ ("*" || "/" || "%"), ExpressionTerm, [Level4Expr]];
+
+
+NotExpression = { "!" }, ExpressionTerm;
+
+ArrayLookupExpression = ExpressionTerm, "[", ExpressionTerm, "]";
```
baziotis: Keep the LHS as just `Identifier` to make it simple but the RHS (i.e. the part inside `[]`) should be `Expression`).

baziotis:
> Currently, expressions such as `Identifier [ true ]` or `new int [ true && false ]` are possible.
> I am of the opinion that rectifying this might make the grammar more unreadable since we could easily have this pop up as an error during the semantic checking phase. What is your opinion on this?

We should definitely not prevent such expressions _grammatically_. As you said, this is part of the semantic analysis.

grammar.ebnf
Doryanni: You mean, `BinaryExpression = Level1Expr, { "&&", Level1Expr}` right?

baziotis: Yes thanks, my bad, I meant to use repetition.

Doryanni:
Array lookup, lvalue, and function call make this grammar ll(2) due to the way they are written. I could group them together, and use `Identifier` as the first token, but that would make no semantical sense.

baziotis:
> Array lookup, lvalue, and function call make this grammar ll(2) due to the way they are written. I could group them together, and use `Identifier` as the first token, but that would make no semantical sense.

Why LL(2)?

**Edit**: Oh wait, yeah it's LL(2) IIRC because by seeing an Identifier, you don't know in what rule you are but it doesn't matter in this case, I think we can handle it just fine in the parsing.

Doryanni:
they all start with `Identifier` and thus, in order for the grammar to not be ambiguous, the token after that is also needed. Thus, the grammar becomes LL(2).

baziotis:
Yeah, just commented it, it doesn't matter, don't worry. There is another place that it _will_ matter and we will have to adjust to the lexer accordingly. Specifically, to know whether we have finished parsing declarations (which will appear at the start of the function) and started statements.
But, we'll discuss that when time comes.

Doryanni:
Also, I propose we make many sub-issues from this point forward, as to give other people the chance to work on this.

baziotis:
Yes, this is a good thought, but probably I will postpone them for now. Essentially, most of the rest of the problems that we will have to solve in the grammar, require to know how we will do parsing. So, I'll wait for then in order for people to be able to understand why certain decisions are made.

grammar.ebnf
```diff
@@ -1,8 +1,15 @@
...
+* Authors:   Ioannis Cheilaris, Stefanos Baziotis
+*)
+
+
 Module = {Declaration}, ? EOI ? ;
 
 Declaration = StructDecl | FunctionDef;
 
-FunctionDef = "func", Identifier, "(", FuncDefParamList, ")", "{", {Statement},
+FunctionDef = "func", Identifier, "(", FuncDefParamList, ")", [":", Type] "{", {Statement},
```
baziotis: Ok, at the start of the function we should have all variable declarations. So, you can make a `VarDeclaration` rule. We can do it by looking the samples. Don't think of it too hard as it's quite possible that we will have to change it. :)

baziotis (commented):
Apart from the little comment, I think it's generally quite ok. I'll have to look it more carefully some time.
But one important thing to remember is that we don't need to think _too_ hard about grammars. Grammars are essentially the blueprint for the parser, but there's no reason to think it through 100% before starting the parser (I would say it's actually counter-productive because implementing the parser will reveal problems easily - problems that are not easy to see just by looking the grammar).
That includes things like worrying too much for LL(1) or LL(2). So, probably this will be merged soon.

grammar.ebnf
Doryanni: Wouldn't it be better if we could declare variables wherever and not just at the start? It seems like a very intense restriction.

baziotis: It depends on what you mean "better". For the programmer writing in DIL, probably yes (although there are a lot of people that prefer the ANSI C style of declarations - i.e. at the start of the function). But for the programmer implementing DIL (i.e. us), no. It's not such a big of a deal, but better leave unneeded complications for now.

baziotis:
Alright, please add a license / copyright notice (look other files for examples).
Then, I think we're mostly done.

Doryanni:
It's at the top of the file in comment form.

baziotis:
Haha I totally missed that mate sorry.

Issue #2, opened 2019-11-25
Function that scans an integer


baziotis:
Write a function that reads integer literals. It should handle base 10, base 8 (octal) and base 16 
(hexadecimal) numbers. It should provide warnings handle erroneous input gracefully.
That includes: out of base digits, overflow.

## Error handling
We want something like `atoi()` but `atoi()` has a bad interface because the function has no way to clearly notify the caller about failure.
A usual idiom in C/C++ is if we have a function that returns a value of a specific type, to pick an invalid value (of this type) that we return to signify that there was an error.
For example say we have a function that is supposed to find a number in an array and return the index of this number. Let's say it returns an `int`. Note that `int` can represent negative values but those are invalid values for an index (array indexes are >= 0).
So, we could pick a specific value out of those invalid values (e.g. `-1`) which we will return to notify the caller that we could not find the number. This is a _contract_ between the caller and callee
and the caller should know about it.

In the same way, functions that are supposed to return a pointer, can use `NULL` to signify failure.

This method of error handling, while good and widely used,
fails when there is _no_ invalid value to pick. In our example, we call a function that reads
an integer value and returns. Well... all possible integer values are valid. What this means is
we have to return 2 values back. One that is the integer value just read and one that holds
if there was an error or not.

Some programming languages provide multiple return value functionality but C / C++ not.
In C / C++ the usual ways to implement this is:
a) By creating a struct that holds those values, for example:
```
typedef read_int_ret {
    int val;
    int error;
} read_int_ret_t;
```
Unfortunately makes the code unnecessarily convoluted by introducing naming all over the place.
So, another way is:
```
int read_int(const char *input, int *val);
```
So, we pass a _pointer_ in which we save the value while what we return signifies the error.

nsoul97:
how should we represent octal number and hexadecimals in the dil language? Should we follow the C way and have 0(zero) and 0x/0X(zero x/X) as prefixes? Can the letters a,b,c,d,e,f in a hex number be both uppper case and lower case?

baziotis:
> how should we represent octal number and hexadecimals in the dil language? Should we follow the C way and have 0(zero) and 0x/0X(zero x/X) as prefixes? Can the letters a,b,c,d,e,f in a hex number be both uppper case and lower case?

Yes, in both the last 2 questions.

nsoul97:
Should there be a third warning if the input string is "","0", "0x" or "0X"? something like empty numerical expression warning

[...]

PR #22, opened 2019-12-02
Integer scan issue

nsoul97:
Issue: https://github.com/ACM-UoA-Student-Chapter/dil/issues/2
int read_int(const char *input,  int *val) implemenation in cpp and declaration in hpp
the integer literal stored in the input string is converted to integer and stored in the val pointer.
if successful read_int returns the success code INT_SCANNED
if not, read int returns the respective error code
warnings: OVERFLOW_ERROR; OUT_OF_BOUNDS_ERROR; EMPTY_NUM_EXPR_ERROR

int_scan.hpp also contains these defined macros for the success code and the error codes.

### DIFF
--- a/int_scan.cpp
+++ b/int_scan.cpp
@@ -29,7 +29,7 @@ numbers have prefix 0x or 0X.
 
 NOTE:: 0 is a decimal number without prefix.
 */
-static int advance(const char *input) {
+static int skip_prefix(const char *input) {
   if (input[0] != '0' || input[1] == '\0')
     return 0;
   if (tolower(input[1]) != 'x')
@@ -44,7 +44,7 @@ Octal numbers have prefix 0, so start_pos=1.
 Hexadecimal numbers have prefix 0x or 0X, so start_pos=2.
 Any other start_pos is invalid.
 */
-static int check_base(int start_pos) {
+static int find_base(int start_pos) {
   switch (start_pos) {
   case 0:
     return 10;
@@ -81,8 +81,8 @@ error codes are defined as:
 The success code is INT_SCANNED.
 */
 int read_int(const char *input, int *val) {
-  int position = advance(input);
-  const int int_base = check_base(position);
+  int position = skip_prefix(input);
+  const int int_base = find_base(position);
   if (input[position] == '\0')
     return EMPTY_NUM_EXPR_ERROR;
   int sum = 0;
--- a/int_scan.cpp
+++ b/int_scan.cpp
@@ -1,6 +1,8 @@
 #include "int_scan.hpp"
 
-enum Base {OCT=8, DEC=10, HEX=16};
+static const int OCT=8;
+static const int DEC=10;
+static const int HEX=16;
 
 /*
 The function returns a member of the Base enumeration, which represents the base of the input string.
@@ -11,7 +13,7 @@ If the input string does not have one of these prefixes, then the input string r
 The function starts looking for the prefix from the index specified by the position pointer.
 The function also returns the position of the next character of the input to be read after the prefix (if it exists).
 */
-static inline Base check_base(const char *input, int *position)
+static inline int check_base(const char *input, int *position)
 {
     *position=0;
     if (input[*position]!='0')
@@ -75,10 +77,10 @@ static inline int base16_char_to_dec(char digit)
 
 /*
 The function returns the integer value of the digit character passed as an argument depending on the digit's base.
-The digit's base is specified by the Base argument of the function
+The digit's base is specified by the num_base argument of the function
 If the digit is out of bounds, the function returns -1.
 */
-static inline int char_to_dec(char digit, Base num_base)
+static inline int char_to_dec(char digit, const int num_base)
 {
     if (num_base==DEC)
         return base10_char_to_dec(digit);
@@ -96,7 +98,7 @@ Otherwise the function returns false.
 This function computes the new value of the numerical sequnence when the next digit of the input string is taken
 into consideration. The new value of the numerical sequnence is stored in the val pointer.
 */
-static inline bool add(int *val, int digit_decval, Base int_base)
+static inline bool add(int *val, int digit_decval, const int int_base)
 {
     *val=(*val)*int_base+digit_decval;
     return *val<0;
@@ -115,7 +117,7 @@ The success code is INT_SCANNED.
 int read_int(const char *input,  int *val)
 {
     int position;
-    Base int_base=check_base(input,&position);
+    const int int_base=check_base(input,&position);
     bool enter=0;
     *val=0;
     for (int i=position;input[i]!='\0';i++)
--- a/int_scan.cpp
+++ b/int_scan.cpp
@@ -1,118 +1,113 @@
 #include "int_scan.hpp"
 #include <ctype.h>
 
-static const int OCT=8;
-static const int DEC=10;
-static const int HEX=16;
+static const int OCT = 8;
+static const int DEC = 10;
+static const int HEX = 16;
 
 /*
-The function returns a member of the Base enumeration, which represents the base of the input string.
-If the input string starts with the character 0 (zero), then the input string represents an octal number.
-If the input string starts with the characters 0x or 0X (zero x/X), then the input string represents a hexadecimal
-number.
-If the input string does not have one of these prefixes, then the input string represents a decimal number.
-The function starts looking for the prefix from the index specified by the position pointer.
-The function also returns the position of the next character of the input to be read after the prefix (if it exists).
+The function returns a member of the Base enumeration, which represents the base
+of the input string. If the input string starts with the character 0 (zero),
+then the input string represents an octal number. If the input string starts
+with the characters 0x or 0X (zero x/X), then the input string represents a
+hexadecimal number. If the input string does not have one of these prefixes,
+then the input string represents a decimal number. The function starts looking
+for the prefix from the index specified by the position pointer. The function
+also returns the position of the next character of the input to be read after
+the prefix (if it exists).
 */
-static int check_base(const char *input, int *position)
-{
-    int pos=0;
-    if (input[pos]!='0')
-    {
-        *position=pos;
-        return DEC;
-    }
+static int check_base(const char *input, int *position) {
+  int pos = 0;
+  if (input[pos] != '0') {
+    *position = pos;
+    return DEC;
+  }
+  pos++;
+  if (tolower(input[pos]) == 'x') {
     pos++;
-    if (tolower(input[pos])=='x')
-    {
-        pos++;
-        *position=pos;
-        return HEX;
-    }
-    *position=pos;
-    return OCT;
+    *position = pos;
+    return HEX;
+  }
+  *position = pos;
+  return OCT;
 }
 
 /*
-The function returns the integer value of the digit character passed as an argument.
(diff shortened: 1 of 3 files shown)


baziotis:
Great work, I like the thought behind the structuring of the code.
Actually, this function won't be used with NUL-terminated strings but it will get part of a string
which is supposed to contain an integer literal.
I wanted to make the issue more approachable so unfortunately, it's also a little misleading.
With this setup, the error codes, while a good idea, are not useful as what we'd like the printing of a warning and recovery from the error.

Let's start by improving the current code and then we can move to the actual version.
To start off, please reference the issue that this PR solves.

int_scan.cpp
```diff
@@ -0,0 +1,133 @@
+#include "int_scan.hpp"
+
+enum Base {OCT=8, DEC=10, HEX=16};
```
baziotis: `Base` is a very generic name and it may be used elsewhere. In this case, you can just have the numeric values instead of an `enum` as it is pretty understandable.

baziotis (commented):
- **Use of `inline`**: The use of `static` is good and many people don't think of it. It's good to make functions that are not supposed to be used elsewhere "private" (to the linker).
Now, `inline` is pretty much useless here. Unfortunately it's a very bad keyword that makes people
think that it refers to the optimization called inlining. For one, "inlining" is not always good and unless one is a pretty hardcore optimizer and expert in the compiler they use, they should leave it to the compiler (for educational purposes, you may want to look at the attribute `always_inline`).
But then second, the keyword `inline` has more or less nothing to do with inlining. It's a [linkage](https://docs.microsoft.com/en-us/cpp/cpp/program-and-linkage-cpp?view=vs-2019) directive to the compiler, which in turn, redirects it to the linker. It tells the compiler that "hey, this will be _defined_ (note: not simply declared) in multiple places. Make sure you use a single instance". That can happen e.g. when a function is defined in a header file.
- **Use of pointers**: Pointers are a pain to deal with for the compiler. We should use them with care. Something like `check_base()` or `add()` are bad code. They are hard to read and they make the compiler work really hard when it tries to optimize them (or not work at all since pointers disable a lot of optimizations depending on the analyses performed). You can transform this code into e.g.:
```cpp
int pos = *position;
... // Do whatever you want with pos
*position = pos;
```
What is more, `add()` could return the result of the sum. Then, the caller could check if the value has overflowed. That's important because `add()` now has to do 2 things instead of one. All that while the only caller should care about overflowing.

- **Format**: Please, use clang format.

int_scan.cpp
```diff
@@ -0,0 +1,133 @@
...
+If the input string does not have one of these prefixes, then the input string represents a decimal number.
+The function starts looking for the prefix from the index specified by the position pointer.
+The function also returns the position of the next character of the input to be read after the prefix (if it exists).
+*/
+static inline Base check_base(const char *input, int *position)
+{
+    *position=0;
+    if (input[*position]!='0')
+        return DEC;
+    else
```
baziotis, line 20: No need to `else` here because above you return. So, if you didn't, you can just continue and avoid nesting. Same in other places in the code.

int_scan.cpp
```diff
@@ -0,0 +1,133 @@
...
+    for (int i=position;input[i]!='\0';i++)
+    {
+        enter=1;
+        int dec_dig=char_to_dec(input[i],int_base);
+        if (dec_dig==-1)
+            return OUT_OF_BOUNDS_ERROR;
+        if (add(val,dec_dig,int_base))
+            return OVERFLOW_ERROR;
+    }
+    if (!enter && int_base!=OCT)        //input string may be the zero literal "0".
```
baziotis: You can put a similar check above the loop and avoid the `enter` altogether. Something like:
```cpp
if (input[position] == '\0' && int_base != OCT) {
  return EMPTY_NUM_EXPR_ERROR;
}
```

baziotis:
Use `tolower()` or something similar to have to deal only with lowercase.

int_scan.cpp
nsoul97: i used the if-else statements instead of using seperate if statements for the code to be more easily understandable. my goal was better readabilty of the code, since with the if-else statement the condition that corresponds to a particular return value was more evident (or at least that's what I thought). Changing these coding segments is easy, but what's your opinion for future reference?

baziotis: That makes sense but also that way, the code implies that you try to avoid something by using the `else`. Moreover, nesting becomes bad for readability as the code gets larger.

nsoul97:
```
static inline int char_to_dec(char digit, Base num_base)
{
    if (num_base==DEC)
        return base10_char_to_dec(digit);
    else if (num_base==OCT)
        return base8_char_to_dec(digit);
    else
        return base16_char_to_dec(digit);
}
```
Should I have an assertion num_base==HEX before the last return statement?

baziotis:
> ```
> static inline int char_to_dec(char digit, Base num_base)
> {
>     if (num_base==DEC)
>         return base10_char_to_dec(digit);
>     else if (num_base==OCT)
>         return base8_char_to_dec(digit);
>     else
>         return base16_char_to_dec(digit);
> }
> ```
> 
> Should I have an assertion num_base==HEX before the last return statement?

Yes. Use `assert`s in abundance. Note that you can comment on code by going to the tab "Files Changed" and clicking on a specific line (or select multiple lines).

int_scan.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+  pos++;
+  if (tolower(input[pos]) == 'x') {
+    pos++;
+    *position = pos;
+    return HEX;
+  }
+  *position = pos;
+  return OCT;
+}
```
baziotis: I think it's better if you transform this to:
```cpp
  int ret;
  int pos = 0;
  if (input[pos] != '0') {
    ret = DEC;
  } else if (tolower(input[pos+1]) == 'x') {
    pos += 2;
    ret = HEX;
  } else {
    ++pos;
    ret = OCT;
  }
  *position = pos;
  return ret;
```
You can even be "smarter" by observing that for every base, there's a different advance in position.
To put it otherwise, every base is identified by its advance. So, this function could be renamed to `advance_pos()`
and the rest of the code work like:
```cpp
int advance = advance_pos(input);
if (advance == 0) // DEC
else if (advance == 1) // OCT
...
input = input + advance;
```
It's a good way to avoid passing a pointer but I'm not sure whether I recommend it :p

int_scan.cpp
```diff
@@ -0,0 +1,115 @@
...
+returns -1.
+*/
+static int char_to_dec(char digit, const int num_base) {
+  if (num_base == DEC)
+    return base10_char_to_dec(digit);
+  if (num_base == OCT)
+    return base8_char_to_dec(digit);
+  assert(num_base == HEX);
+  return base16_char_to_dec(digit);
+}
```
baziotis, line 71: This function could use a map that maps digits to their values. Something like:
```cpp
uint8_t char_to_digit[256] = {
    ['0'] = 0,
    ['1'] = 1,
    ['2'] = 2,
    ['3'] = 3,
    ['4'] = 4,
    ['5'] = 5,
    ['6'] = 6,
    ['7'] = 7,
    ['8'] = 8,
    ['9'] = 9,
    ['a'] = 10, ['A'] = 10,
    ['b'] = 11, ['B'] = 11,
    ['c'] = 12, ['C'] = 12,
    ['d'] = 13, ['D'] = 13,
    ['e'] = 14, ['E'] = 14,
    ['f'] = 15, ['F'] = 15,
};
```
Then, you can simply do:
```cpp
int val = char_to_digit[digit];
if ((val == 0 && digit != 0) || val >= num_base)
  return -1;
return val;
```

nsoul97: That was a cool idea. First time i saw designated initialization syntax so i looked it up. It turns out that designated initialization for arrays is not valid for c++ as opposed to c.
`https://en.cppreference.com/w/cpp/language/aggregate_initialization`

baziotis: You're right. Well, they're supported in C++20 but we don't want to be that progressive..
In the `main()` you're testing that, just initialize the array. This function is essentially going to be part of the lexer and the lexer will need a general initialization function anyway. So, we will throw this initialization there.

_Note_: You can use the syntax: `[link text](link url)` to post links. Alternatively, you can highlight the the text you want to act as link and use the icon near the `<>` above.

nsoul97: Should the array be declared globally or locally inside the main function and then passed as an argument to the read_int function?

int_scan.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef __INT_SCAN_HPP
+#define __INT_SCAN_HPP
+
+#define INT_SCANNED 0;
+#define OVERFLOW_ERROR -1;
+#define OUT_OF_BOUNDS_ERROR -2;
+#define EMPTY_NUM_EXPR_ERROR -3;
```
nsoul97: Do you think I should keep these macros or there may be conflicts later in the implementation?

int_scan.cpp
baziotis: Globally, because the `main` function will act as a "filler" for the array.

int_scan.hpp
baziotis: As I've said, this is not the final form of the function. Just trying to make it somewhat simple and generic before moving on to a more specific usage.
In this more specific form, these won't be used, because the function won't return an error, but instead will print warnings. There are reasons for this that I will explain when we move to implement the function in its final form.

nsoul97:
Should I upload a main with various test cases checked by assert or we will use a unit testing framework?

baziotis:
The former for now.

baziotis (commented):
Great code! Just one proposed change and I think it's ready to be merged.

int_scan.cpp
```diff
@@ -0,0 +1,99 @@
...
+its value in the val pointer. This function returns the respective error code in
+case of error or the success code when the integer is successfully scanned. Ther
+error codes are defined as:
+-->OVERFLOW_ERROR
+-->OUT_OF_BOUNDS_ERROR
+-->EMPTY_NUM_EXPR_ERROR
+The success code is INT_SCANNED.
+*/
+int read_int(const char *input, int *val) {
+  int position = advance(input);
```
baziotis: You might add a comment above here like "`position` is where the actual digits start (having consumed any possible prefix)".

nsoul97: I have actualy wrote that at the description of the advance function.
Do you think I should add the comment again when I use the function?

baziotis: Indeed.. Here's a better solution. Name `advance` -> `skip_prefix` which justifies that it returns a position and also makes clear what it does in the code without a comment.
Related to that, it'd be better to name `check_base` -> `find_base`, because you don't check something, rather, you find something by passing the `position`.

baziotis (commented):
Alright, that's ready to be merged. That's not the actual function that we will use in the end, but it's in a pretty good form to be refactored to the final version. A new issue will be opened then.

As a final note, when you resolve something that was mentioned in a specific comment, you may then
hit "Resolve conversation" so that the reviewer doesn't have to check which comments are still unresolved.

baziotis:
My bad, I forgot the license / copyright addition. Please, add a copyright / license statement on top of the 3 files that you added, with this template:
```cpp
/*
* Authors:   <YOUR_NAME>
*/
```

Issue #9, opened 2019-11-25
Colored error messages


baziotis:
Write a variadic template function that displays syntax error messages. You might not know what template or variadic means but trust me, if you do want to learn, this is a very easy issue.
We basically want a function that we can e.g. write: `syntax_error("Invalid character literal");`
And get `Syntax Error: Invalid character literal`. But, we want the `Syntax Error:` part to
be **colored**. Preferably red. :)

Still though I haven't explained why do we need the function to be a variadic template function.
Well, we need to do things like this:
```
const char *id = ...;
...
syntax_error("Type with id: ", id, " is invalid.");
```
So, note that we pass multiple typed parameters in the function. And we want to pass
as many as we want. Those 2 together most of the time mean variadic templates.
The function should be able to recognize that `id` is of type `const char *` and print it
accordingly.

But that is maybe not the easiest thing in the world. You may want firstly to start with a function
that just prints one string. Then, multiple strings (with variadic templates), then try to recognize
different types etc.

billsioros:
I would like to address this issue. Is there anything else that I should bare in mind while implementing this feature, like avoiding the usage of iostream e.t.c ?

baziotis:
> I would like to address this issue.

Ok, you got an assignment.
> Is there anything else that I should bare in mind while implementing this feature, like avoiding the usage of iostream e.t.c ?

Yeah, better just use `printf`. Unless there is something quite useful that `iostream` provides that
`printf` doesn't and I haven't thought about it. Now that I think of it, with `iostream` one would probably do it very differently than what I have in mind (this different way is not good IMO). Let's not spoil it, feel free to propose ideas with or without `iostream` (or whatever you want).

billsioros:
I have come up with a solution that uses recursive calls and I would like to get rid of them and make everything a single statement.

One solution to this issue would be using the C Preprocessor, while another one would be implementing a constexpr string.

Is any one of the two more preferable?

Edit: The C Preprocessor is considered by many a bad practise but I consider it simply a tool, while constexpr string results in memory consumption and makes the implementation a tiny bit more complicated.

baziotis:
> I have come up with a solution that uses recursive calls and I would like to get rid of them and make everything a single statement.
> 
> One solution to this issue would be using the C Preprocessor, while another one would be implementing a constexpr string.
> 
> Is any one of the two more preferable?
> 
[...]

PR #23, opened 2019-12-03
Variadic Template Error Logging Initial

billsioros:
Addressing issue #9 

The implementation supports some basic types and it can be easily extended to support more in case that is needed.

One utility file named panic.hpp is included that is responsible to handle errors. Feel free to remove it / replace it.

ATM the coloring of the text is based on UBUNTU/LINUX, and MacOS ONLY terminal color codes.


baziotis (commented):
Impressive but this is too complicated for the problem it tries to solve. :) Why not use simple function overloading?
Also, no need for these namespaces.

billsioros:
> Impressive but this is too complicated for the problem it tries to solve. :) Why not use simple function overloading?

Touché

> Also, no need for these namespaces.

I added the outter namespace so that I don' t pollute the global namespace. The inner namespace serves the purpose of anyone using this module not concerning themselves with implementation details for example when autocompletion in an IDE / text editor occurs.

Now if you insist they are removed, thy will be done.

baziotis:
> > Also, no need for these namespaces.
> 
> I added the outter namespace so that I don' t pollute the global namespace. The inner namespace serves the purpose of anyone using this module not concerning themselves with implementation details for example when autocompletion in an IDE / text editor occurs.
> 
> Now if you insist they are removed, thy will be done.

Trust me, they do more pollution to the code than the global namespace. :p

billsioros:
> Trust me, they do more pollution to the code than the global namespace. :p

Ok done. I inlined all the functions as they are of small code size. If a source file is prefered from the header-only approach I' ll change it asap.

log.hpp
```diff
@@ -0,0 +1,58 @@
...
+#pragma once
+
+#include <utility>
+#include <cstdio>
+#include <cstdlib>
+#include "panic.hpp"
+
+inline void single_print(const char * arg, const char * next)
+{
+    panic(std::printf("%s%s", arg, next) < 0);
```
baziotis: You can just `assert` instead of `panic()`. Why introduce a new thing for one to look up? :)
But in this case, we can safely remove it. It's not that a `printf` can never fail. It's that we don't care that much.

baziotis (commented):
It's getting better, but you can simplify it even more! No need for `next` or `std::forward` or `constexpr`.

baziotis:
> Ok done. I inlined all the functions as they are of small code size. If a source file is prefered from the header-only approach I' ll change it asap.

No, I think that was a good initiative.

billsioros:
> It's getting better, but you can simplify it even more! No need for `next` or `std::forward` or `constexpr`.

Shouldn' t **std::forward** be used when passing the arguements of a template function to another so that they are treated correctly as either l or rvalues ?

Also  doesn' t **constexpr** in this context result in code being eliminated at compile time as the size of a variadic template is known at compile time ?

Edit: I know I could call printf with desired "next" (i.e. " " or "\n") but this results in multiple function calls and that' s the reason i introduced **next**.

baziotis:
> > It's getting better, but you can simplify it even more! No need for `next` or `std::forward` or `constexpr`.
> 
> Shouldn' t std::forward be used when passing the arguements of a template function to another so that they are treated correctly as either l or rvalues ?
>

Yes, and you pointed me to something I missed: No need to pass references. Why complicate the code when all the values you print are trivial and can be passed by value?

> Also the constexpr in this context doesn' t result in code being eliminated at compile time as the size of a variadic template is known at compile time ?

Yes, but I'm thinking of a solution that you don't need this check at all.
As a note, this expression is simple enough that the compiler will eliminate it itself.
Even in such cases though, I'm positive to `constexpr` because it makes clear that it should
be executed at compile-time.

billsioros:
I guess [this](https://github.com/ACM-UoA-Student-Chapter/dil/pull/23/commits/3792fed390baede193219aba5f03529a06230851) is what you are talking about.

I wanted to avoid doing that because of that extra printf statement used to print the new line. It' s not graceful IMO but it' ll most probably perform better than the initial solution.

baziotis:
> I guess [this](https://github.com/ACM-UoA-Student-Chapter/dil/pull/23/commits/3792fed390baede193219aba5f03529a06230851) is what you are talking about.
> 

Yep, but _a little_ bit simpler. See comments in the code.
> I wanted to avoid doing that because of that extra printf statement used to print the new line. It' s not graceful IMO but it' ll most probably perform better than the initial solution.

We don't want to sacrifice that much readability for one `printf` that is anyway going to be separated because sometimes we won't want to print a newline.

log.hpp
```diff
@@ -0,0 +1,54 @@
...
+inline void print(Last last)
+{
+    single_print(last); std::printf("\n");
+}
+
+template <typename First, typename Second, typename ...Rest>
+inline void print(First first, Second second, Rest... rest)
+{
+    single_print(first); print(second, rest...);
+}
```
baziotis: You can just write:
```cpp
template <typename First, typename ...Rest>
inline void print(First first, Rest... rest) {
  print(first);
  print(rest);
}
```
and replacing `single_print()` with `print()`.

log.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+template <typename ...Args>
+inline void syntax_error(Args&&... args)
+{
+    print("\033[31mSyntax Error:\033[0m", std::forward<Args>(args)...);
```
baziotis: Add doc on what these weird octal numbers do and also remove refs.

log.hpp
billsioros: Yeah if the newline printing is going to get separated from the rest of the function this is understandable.

baziotis (commented):
Use `clang-format` and add license on the top of the file as follows:
```cpp
/*
* Authors:   <YOUR_NAME>
*/
```

log.hpp
baziotis: Even if it wasn't, it's not worth it. That's a personal opinion of course and in performance circumastances that I've not seen that might matter. But generally, I think that in big projects it's a bad idea to sacrifice so much readability for such a little perf benefit.

billsioros: Of course, I understand. What I meant was that if you got just a single separator it makes things much simpler.

Edit: I got confused when you mentioned function overloading before, as now we are talking about template specialization (if I' m not mistaken) so that' s why i went with that xD

log.hpp
billsioros: Of course :+1:

log.hpp
baziotis: No, AFAIK (and I don't pretend to be a C++ guru, neither would I ever waste time to become one) this is function overloading. Template specialization is what you did before.

billsioros:
Everything should be fine now :D

baziotis (commented):
Wait, why did we use again template specialization?

billsioros:
> Wait, why did we use again template specialization?

Based on your reply I arrived at that conclusion. You are using a print with a single parameter and a parameter pack, so I assumed that you were indicating template specialization of the base case.

```cpp
template <typename First, typename ...Rest>
inline void print(First first, Rest... rest) {
  print(first);
  print(rest...);
}
```

baziotis:
> > Wait, why did we use again template specialization?
> 
> Based on your reply I arrived at that conclusion, because you are using a print with a single parameter and a parameter pack, I assumed you were pointing towards template specialization of the base case.
> 
> ```c++
> template <typename First, typename ...Rest>
> inline void print(First first, Rest... rest) {
>   print(first);
>   print(rest...);
> }
> ```

Yep, but you don't need template specialization in the base case. ;)
Just normal function overloading, i.e.:
```cpp
inline void print(const char *s) {
  print("%s ", s);
}
... // other overloads
template <typename First, typename ...Rest>
inline void print(First first, Rest... rest) {
  print(first);
  print(rest...);
}
```

baziotis:
Any news in this?

billsioros:
I've started an internship and it completely slipped my mind. I'm really sorry. I'll make the final changes till Sunday.

baziotis:
> I've started an internship and it completely slipped my mind. I'm really sorry. I'll make the final changes till Sunday.

No worries and best of luck with the internship! Ok, looking forward to them.

billsioros:
Everything should be OK now !

PR #24, opened 2019-12-04
The read_int issue

TheodoreAlenas:
- The coding style matches the default in the project
  (int and not bool, no camelCase, few spaces etc)

- Three functions come in a package, the wanted and two that help it
  (a read_int for a single character and a base_of)

- Useful and short comments in both the header and the cpp file

- A good balance between performance and readability

### DIFF
--- a/DIL ACM/read_int.cpp
+++ b/DIL ACM/read_int.cpp
@@ -1,0 +1,99 @@
+// info about me at the bottom
+#include "read_int.h"
+#include <cctype>
+using std::isdigit;
+
+int base_of(const char *token) {
+
+  if (token[0] == '\0') { // error
+    return 0;
+  }
+
+  if (token[0] == '0') {
+
+    if (token[1] == '\0') { // token is "0"?
+      return 10;
+
+    } else if (token[1] == 'x') { // dexahecimal
+      return 16;
+
+    } else if (token[1] == 'b') { // binary
+      return 2;
+
+    } else { // octal
+      return 8;
+    }
+  } else {
+    if (isdigit(token[0]))
+      return 10; // decimal
+  }
+
+  return 0; // error
+}
+
+int read_int(char c, const int base) {
+
+  int number = -1;
+  // an invalid value that should change
+
+  if (base == 16) { // base 16
+
+    number = read_int(c, 10);
+    if (number != -1)
+      return number;
+
+    if ('a' <= c && c <= 'f')
+      return c - 'a' + 10;
+
+    if ('A' <= c && c <= 'F')
+      return c - 'A' + 10;
+
+    return -1;
+  } 
+  // any other base
+  number = c - '0';
+  if (number < 0 || number >= base)
+    return -1;
+
+  return number;
+}
+
+int read_int(const char *token, int *number) {
+
+  *number = 0;
+
+  int base = base_of(token);
+  if (!base)
+    return 0;
+  // moving after the "0b.." or "0x..." ("0.." is fine)
+  if (base == 2 || base == 16)
+    token += 2;
+
+  int digit;
+
+  // the pointer 'token' is pushed to the end
+  while (*token != '\0') {
+    digit = read_int(*token, base);
+    if (digit == -1)
+      return 0; // invalid character as digit
+
+    *number *= base;
+    *number += digit;
+    if (*number < 0)
+      return 0; // overflow
+
+    ++token;
+  }
+
+  return 1; // the number is fine
+}
+
+/*
+
+name:			Theodoros Dimakopoulos
+GitHub:		TheodoreAlenas (nickname is Alenas)
+e-mail:		dimakopt732@gmail.com
+place:		Greece, UOA (known there as sdi1900048)
+last edit:	4-12-19
+
+*/
--- a/DIL ACM/read_int.h
+++ b/DIL ACM/read_int.h
@@ -1,0 +1,22 @@
+#pragma once
+
+// takes the token. Returns the base, or 0 if error
+int base_of(const char *);
+
+// takes ONE character and the base. returns -1 on error
+int read_int(char, const int);
+
+// takes the token and a variable to fill
+// with the integer. On error returns 0.
+int read_int(const char *, int *);
+
+
+/*
+
+name:			Theodoros Dimakopoulos
+GitHub:		TheodoreAlenas
+e-mail:		dimakopt732@gmail.com			
+place:		Greece, UOA (known there as sdi1900048)
+last edit:	4-12-19
+
+*/


TheodoreAlenas:
I didn't know that nsoul77 had already finished his code. In fact, I find his version to be much more elaborate.

baziotis:
> I didn't know that nsoul77 had already finished his code. In fact, I find his version to be much more elaborate.

That's ok, no problem. In general, check the issue if there is any PR references (for example, if you see the issue that you're trying to solve, you'll see Nikos' PR at the bottom).
Other than that, check if it has any assignees (this does not). And finally, check the PRs in general.

Other than that, I can still provide a review if you want for educational purposes. This is not something that happens generally in projects for obvious reasons (duplicate work). But since 
this is an educational project, I think that such things are beneficial.

TheodoreAlenas:
Thanks for thr reply! Of course I would love even a draft critique, be it whatever.
Sorry for any obvious mistakes I make, I am a little overwhelmed with GitHub.

baziotis:
> Thanks for thr reply! Of course I would love even a draft critique, be it whatever.
> Sorry for any obvious mistakes I make, I am a little overwhelmed with GitHub.

That's ok, this project was created for people to experiment!

DIL ACM/read_int.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/*
+
+name:			Theodoros Dimakopoulos
+GitHub:		TheodoreAlenas (nickname is Alenas)
+e-mail:		dimakopt732@gmail.com
+place:		Greece, UOA (known there as sdi1900048)
+last edit:	4-12-19
+
```
baziotis, line 98: What's the rationale behind those?

DIL ACM/read_int.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+// takes the token. Returns the base, or 0 if error
+int base_of(const char *);
+
+// takes ONE character and the base. returns -1 on error
+int read_int(char, const int);
+
+// takes the token and a variable to fill
+// with the integer. On error returns 0.
```
baziotis, line 10: What integer?

baziotis: To be more specific: In places like this, you don't want documentation that assumes that the reader knows what you're trying to do. For example, here "fill with the integer" assumes that the reader knows about what integer you're talking about. Also, the "token" is a bad term here, it means other things in compiler parlance and not what this function takes here.

A better doc could be e.g. (rename `token` to `input`): Extract the integer value from the integer literal found in `input` (assume that `input` starts with such a literal). Put the value in <second_param_name>.

DIL ACM/read_int.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+// takes the token. Returns the base, or 0 if error
+int base_of(const char *);
+
+// takes ONE character and the base. returns -1 on error
+int read_int(char, const int);
```
baziotis, line 7: These functions are not supposed to be used by the user and hence should be "private", i.e. `static` in the file they're used.

DIL ACM/read_int.cpp
```diff
@@ -0,0 +1,99 @@
...
+    if (digit == -1)
+      return 0; // invalid character as digit
+
+    *number *= base;
+    *number += digit;
+    if (*number < 0)
+      return 0; // overflow
+
+    ++token;
+  }
```
baziotis, line 86: Please, don't use `*number` everywhere. This is bad for multiple reasons, one of which is readability.
You could remedy this by doing e.g.:
```cpp
int num;
...  // do whatever you do in *number with num
*number = num;
```

TheodoreAlenas:
Alright, that was very informative! Just a few things though :

DIL ACM/read_int.h
TheodoreAlenas: Shall I declare them in the cpp file? Is that what you mean?

TheodoreAlenas:
How should I make a declaration static? Do you mean move it to the cpp file?

And would it be wise to use "#define num *number"?

baziotis:
> How should I make a declaration static? Do you mean move it to the cpp file?
> 
> And would it be wise to use "#define num *number"?

When your questions are about a specific comment that was made in your code, it's easier to reply there. That's better for you, the reviewer and anyone reading this PR. :) Just like you did with the question "Shall I declare...".

So, please do me a favor and move your comments there and I'll answer asap.

DIL ACM/read_int.cpp
TheodoreAlenas: Is "#define num *number" a good idea? For readability, too, because "int num" could mean making a temporary variable.

TheodoreAlenas:
Done!

DIL ACM/read_int.cpp
TheodoreAlenas: They seem useful. Does GitHub make up for them?

DIL ACM/read_int.cpp
baziotis: Oh no, it's a very bad idea. :) We could have a very long discussion about the preprocessor and its usages but for the sake of brevity (and not boring people), I'll try to give you some useful thoughts (if you actually care about this discussion, you may send me PM in Slack):

To start off, the preprocessor and C are 2 languages that don't communicate very well. The one language does not know about the other. Hence, the preprocessor should be used with great care.

Now, In this specific case, using this idiom introduces a couple of bad things. First, if you write `num` anywhere in your code, to the reader _it seems_ like a variable, but it's not. That's why you usually see preprocessor entities having very distinctive names (e.g. all caps). To make clear that it's some preprocessor magic.
Even if you did give it a very distinctive name, like `NUM`, now you have introduced other bad things. First, usually preprocessor entities are about constant things (e.g. `#define NUM 10`). So, the reader will probably assume that this is a constant of some kind. Then second, you have hidden the fact that you're using a pointer under the hood. You _don't_ want to hide this fact from the reader.
Third, think that someone reads this code:
```cpp
int *number = ...;
num = 2;
```
or whatever, `number` could be a function param as in our example, e.g.:
```cpp
void some_function(int *number) {
  .. // some code
  num = 2;
}
```
Basically, what I'm trying to say is that the user sees `number` at one point and then some usage of `num`. By looking at it, those seem unrelated, however, they are very related. :P 
This is hidden from the user, again making this code very hard to read and predict.

In any case, it would always be better and more readable to write `*number` instead of replacing it textually (because that's essentially what you do with this `#define` - a search and replace for `*number` with `num`). Not to mention that the `*number` might be used somewhere else in the code which is even worse, as you may not intend to replace it there.
That's why you'll see people `#undef` stuff when they don't need them to avoid such things.

Essentially, what you want is to just _not_ use a pointer. Hence, you introduce a temporary _local_ variable, do whatever you want and you just write its value to the pointer once in the end. As I said:
```cpp
int num = ...;
.. // blah blah do things with num that you would do with *number
*number = num;  // don't forget to write the value in the end.
```
That's better for you, the reader and the compiler.

> For readability, too, because "int num" could mean making a temporary variable.

Why is that bad for readability?

DIL ACM/read_int.h
baziotis: I mean first of all to remove them from here. In the `.h` files you put functions that are supposed to be used by the user. To implement something, you may use _more_ functions than those that are supposed to be used by the user. For example here, the user is supposed to only call (and care) for `read_int(const char *, int *);`. But, to implement this function, you have defined other functions, that the user does not need to know about. So, you don't declare them here.
What you do is you _define_ them in the cpp file. You don't necessarily need to declare them if they're defined before they're used (tell if you're confused about the difference of a definition and a declaration).
Now, the extra thing that you do is that you define the function `static`, e.g.:
`static int base_of(const char *);`. In that way, this function can't be called from another source file (this is a very simplistic and almost incorrect description, but I don't want to make it complicated at this point :) ).

DIL ACM/read_int.cpp
baziotis: Yes. To be more specific, _Git_ makes up for them, but there's one thing you should include: **Copyright / License**. You don't need to include it in this PR, because it won't be merged. But, generally, the project in which you're working should have a license, and you should add your copyright if you've done significant changes (apart from a legal issue it's also an ethical one).

Issue #3, opened 2019-11-25
Provide the list of tokens as an enum


baziotis:
Provide an `enum class` that will help us identify tokens.

Tokens are the smallest piece of information that the compiler understands.
For example, the characters `in` in C mean nothing. But the characters `int`
mean something - they're one token signifying the keyword `int`.
In the same way a number e.g. `102` is an integer literal token.
We want the list of all possible tokens of DIL.
_Important note_: We want all the possible _categories_ of tokens. That is,
102 and 103 are different tokens but they belong to the same category: integer literals.

You can start off like this:

```
enum class TOK {
    UNDEFINED,  // token of undefined kind (i.e. error)
    EOI,                 // end of input
    INTLIT,             // integer literal
    ...
};
```
The structure we'll use to save tokens we'll have a field of type `TOK` named e.g. `kind`.
Then we can distinguish tokens as: 
```
if (kind == TOK::INT_LIT) {
    // Do something specific for integer literals
}
```

nsoul97:
I'll try this out.

baziotis:
So, assigning?

nsoul97:
So some questions about DIL.
Which of the following will DIL support?
1)switch, case statements
2)const
3)continue, break
4)'++', '--'
5)assignment operators other than '=' (e.g. '+=', '-=', '<<=')
6)ternary operator('?:')
7)bitwise operators (e.g. '<<', '|', '~')

baziotis:
None, at least for now. For the time being, you can assume that all we will use is what is found in the samples.

nsoul97:
Is EOI DIL's EOF or does it represent the ; token at the end of a statement?

baziotis:
The former. `;` is a separate token.

nsoul97:
I found that comments in C are stripped by the preprocessor so they are not considered to be tokens. What about DIL? Will we use a preprocessor before the compilation? In that case `//`, `/*`, `*\` are not tokens, right?

baziotis:
Note that you can write code like `//`, `/*`, `*\` (you can hit edit in my comment to see).
So yes, in C they're usually removed by the preproc although this is not necessary.
DIL won't have a preproc for various reasons, so the comments should be stripped off by the lexer,
the first part of the compiler. Note that `//`, `/*`, `*/` are **_not_** tokens. A token is the least possible piece of information (or entity) that has meaning to a compiler.
By definition, comments have no meaning to the compiler.
[...]

PR #27, opened 2019-12-09
Tokens enum

nsoul97:
Issue: #3 
First go with the tokens enum class.
The enumerations in the enum are supposed to support the .dil sample source files.
(e.g. string keyword not included yet, since it is not used in is_prime.dil and quicksort.dil files)

### DIFF
--- a/tokens_enum.hpp
+++ b/tokens_enum.hpp
@@ -17,10 +17,9 @@ enum class TOK {
   UNDEFINED,   // token of undefined kind (i.e. error)
   EOI,         // end of input
   INT_LIT,     // integer literal
-  MAIN,        // main keyword
   INT,         // int keyword
   BOOL,        // bool keyword
-  FUNCTION,    // func keyword
+  FUNC,        // func keyword
   RETURN,      // return keyword
   IF,          // if keyword
   WHILE,       // while keyword
@@ -29,25 +28,25 @@ enum class TOK {
   NEW,         // new keyword
   PLUS,        // + operator
   MINUS,       // - operator
-  MULT,        // * operator
+  MUL,         // * operator
   DIV,         // / operator
   MOD,         // % operator
   ASSIGN,      // = operator
-  EQUAL,       // == operator
-  UNEQUAL,     // != operator
-  GREATER,     // > operator
-  GREQ,        // >= operator
-  LESS,        // < operator
+  EQ,          // == operator
+  NEQ,         // != operator
+  GT,          // > operator
+  GEQ,         // >= operator
+  LT,          // < operator
   LEQ,         // <= operator
   COLON,       // : token
   COMMA,       // , token
   ID,          // function (e.g. quicksort) or variable (e.g. len) identifiers
-  START_ARRAY, // [ token
-  END_ARRAY,   // ] token
-  START_PAREN, // ( token
-  END_PAREN,   // ) token
-  START_SCOPE, // { token
-  END_SCOPE,   // } token
+  LBRACKET,    // [ token
+  RBRACKET,    // ] token
+  LPAR,        // ( token
+  RPAR,        // ) token
+  LBRACE,      // { token
+  RBRACE,      // } token
   SEMICOLON    // ; token
 };
 
--- a/tokens_enum.hpp
+++ b/tokens_enum.hpp
@@ -14,40 +14,40 @@ Therefore to use an enumerator name we have to use the scope resolution
 operator. e.g. TOK::UNDEFINED
 */
 enum class TOK {
-  UNDEFINED,   // token of undefined kind (i.e. error)
-  EOI,         // end of input
-  INT_LIT,     // integer literal
-  INT,         // int keyword
-  BOOL,        // bool keyword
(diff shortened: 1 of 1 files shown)


tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+operator. e.g. TOK::UNDEFINED
+*/
+enum class TOK {
+  UNDEFINED,   // token of undefined kind (i.e. error)
+  EOI,         // end of input
+  INT_LIT,     // integer literal
+  MAIN,        // main keyword
+  INT,         // int keyword
+  BOOL,        // bool keyword
+  FUNCTION,    // func keyword
```
baziotis: `FUNCTION` -> `FUNC`

tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+  RETURN,      // return keyword
+  IF,          // if keyword
+  WHILE,       // while keyword
+  FALSE,       // false keyword
+  TRUE,        // true keyword
+  NEW,         // new keyword
+  PLUS,        // + operator
+  MINUS,       // - operator
+  MULT,        // * operator
```
baziotis: `MULT` -> `MUL`

tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+  DIV,         // / operator
+  MOD,         // % operator
+  ASSIGN,      // = operator
+  EQUAL,       // == operator
+  UNEQUAL,     // != operator
```
baziotis: `EQUAL` -> `EQ`
`UNEQUAL` -> `NEQ`

tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+  GREATER,     // > operator
+  GREQ,        // >= operator
```
baziotis: `GREATER` -> `GT`
`GREQ` -> `GEQ`

tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+  LESS,        // < operator
```
baziotis: `LESS` -> `LT`

tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+  LEQ,         // <= operator
+  COLON,       // : token
+  COMMA,       // , token
+  ID,          // function (e.g. quicksort) or variable (e.g. len) identifiers
+  START_ARRAY, // [ token
+  END_ARRAY,   // ] token
+  START_PAREN, // ( token
+  END_PAREN,   // ) token
+  START_SCOPE, // { token
+  END_SCOPE,   // } token
```
baziotis: `*_ARRAY` -> `LBRACKET`, `RBRACKET`
`*_PAREN` -> `LPAR`, `RPAR`
`*_SCOPE` -> `LBRACE`, `RBRACE`

tokens_enum.hpp
```diff
@@ -0,0 +1,54 @@
...
+Enum class TOK is used to help identify the tokens of DIL language.
+Since this is an enum class, enumerator names are local to the TOK enum.
+Therefore to use an enumerator name we have to use the scope resolution
+operator. e.g. TOK::UNDEFINED
+*/
+enum class TOK {
+  UNDEFINED,   // token of undefined kind (i.e. error)
+  EOI,         // end of input
+  INT_LIT,     // integer literal
+  MAIN,        // main keyword
```
baziotis: `MAIN` is not a special keyword, it's an identifier. The `main` function is just like any other function (with a difference that it has to exist but this does not make its name a special keyword).

baziotis (commented):
Great work, note that some of the renaming I requested is a matter of preference. Other (like `START_ARRAY`) not so much, meaning, no compiler I know of uses this convention.

nsoul97:
thanks for the notes. will update asap

baziotis:
When you do changes proposed in a specific comment, you can then hit "Resolve Conversation" so that the reviewer doesn't have to search through all of them and try to find which ones were resolved and which not. :)

baziotis (commented):
Add `VOID`, `ELSE`, `DOT`, `NOT` (i.e. `!` operator).

nsoul97:
Where will the tokens `void`, `.` be used?
Should I also add AND, OR enumerations (`&&`, `||` operators respectively)?

baziotis:
> Where will the tokens `void`, `.` be used?

Well, actually we don't need `void`. I was thinking it would be used in the return type
but actually, with the current syntax, you denote that a function returns nothing
by just not writing a return type.
`.` will be used in struct field dereference. Like in C:
```cpp
struct A a;
a.b = 1;
```
And btw, you can add `STRUCT` keyword as well.

> Should I also add AND, OR enumerations (`&&`, `||` operators respectively)?

Yes, but add them as `AND_AND`, `OR_OR` because at some poiint we may choose to support bitwise operators.

baziotis:
Alright, that seems ok. In any case, we don't need to think hard about tokens for now. It seems "sort of ok" so it's good. If we need to add any other token, we'll do it when time comes.

Issue #8, opened 2019-11-25
Character map


baziotis:
Create a character map that will help us check if a character belongs to a certain category.
For example, it will help us check if a character is an octal digit. Or if a character can appear in the middle of an identifier.

More specifically, we need to be able to check if the character:
- is blank (space, tab etc.)
- can appear in the middle of an identifier.
- is an octal digit.
- is a hexadecimal digit.
- can appear at all in the a DIL program.

That last one may be unfamiliar to most. Try to include the character `@` in any place in a
C program and compile. The character `@`, maybe interestingly, can't appear in a C program (that
doesn't mean it can't appear in a DIL program).

This issue deliberately has a vague description. I think it's interesting to discuss what the best
way to achieve the goal is.

TheodoreAlenas:
{'+', '-', '*', X, 'a', 'z', '0', '9', '\0'}

could mean "I contain +-* and from a to z and from 0 to 9".
This is practical because hexadecimal digits would be
{X, '0', '9', 'a', 'f', 'A', 'F', '\0'}
all in all.

X here is a character value noone will ever use, such as 0x0 or 0x1.

So we could make such arrays in a header file, as well as a function that helps identify if a character is "included" in such an array.

By the way, I don't know much about maps but STL maps in C++ don't seem relative to this problem.

baziotis:
TBH, I didn't really understand what you're describing. You just want a way to query a data structure with a character and see if it is of a specific kind. Conceptually something like:
```cpp
char c = ...;
if (map.is_octal(c)) { ... // blah }
```
Let me re-iterate that this is the _conceptual_ idea. What you have to do is much simpler.

nsoul97:
@TheodoreAlenas have you started this issue?

TheodoreAlenas:
No no, of course not. I barely understood it. 😄

baziotis:
> No no, of course not. I barely understood it.

If there's something you don't completely understand at first but you are willing to understand it, please request more info. It's totally acceptable for you to not understand something and for me to not have explained something the best way possible.

You might want to start with another question. Let's say I'm requesting a program from you that should behave as follows. I give you the list of all the ASCII characters (which are 256). For each one, I have a number next to it, like (note that I use the decimal value of each character and not how it's printed, 0 here means the character with decimal value zero (which is the NUL character) and not the _digit_ 0 (which has decimal value 48, as you can see in the [ascii table](https://www.asciitable.com/)):
```
[...]

PR #28, opened 2019-12-10
The character map, as an array

TheodoreAlenas:
Issue https://github.com/ACM-UoA-Student-Chapter/dil/issues/8

The character map that stores the properties of the ASCII characters, such as whether they represent digits.
No namespaces were used and the file management is unorthodox but simple.


TheodoreAlenas:
If I have made basic mistakes on using git, it is because I messed up and I tried to un-mess everything.

baziotis:
> If I have made basic mistakes on using git, it is because I messed up and I tried to un-mess everything.

Seems ok. :) Just add the issue you solve in the description of the PR. For example "Issue: THE_ISSUE_LINK".

character_map.h
```diff
@@ -0,0 +1,91 @@
...
+  for (int i = 33; i < 127; ++i)
+    ch_map[i] = CM_DIL_CHAR;
+
+  // characters allowed in identifiers
+  for (int i = '0'; i != '9'; ++i)
+    ch_map[i] |= CM_IDENTIFIER;
+  for (int i = 'a'; i != 'z'; ++i)
+    ch_map[i] |= CM_IDENTIFIER;
+  for (int i = 'A'; i != 'Z'; ++i)
+    ch_map[i] |= CM_IDENTIFIER;
```
baziotis: You can collapse these 2 loops with:
```cpp
for (int i = 'A'; i != 'Z'; ++i) {
  ch_map[i] |= CM_IDENTIFIER;
  ch_map[tolower(i)] = CM_IDENTIFIER;
```

character_map.h
```diff
@@ -0,0 +1,91 @@
...
+  ch_map['0'] |= CM_OCTAL;
+  ch_map['1'] |= CM_OCTAL;
+  ch_map['2'] |= CM_OCTAL;
+  ch_map['3'] |= CM_OCTAL;
+  ch_map['4'] |= CM_OCTAL;
+  ch_map['5'] |= CM_OCTAL;
+  ch_map['6'] |= CM_OCTAL;
+  ch_map['7'] |= CM_OCTAL;
+  ch_map['8'] |= CM_DECIMAL;
+  ch_map['9'] |= CM_DECIMAL;
```
baziotis: Loop. Same below. ~~Also, `0 -- 7` are also decimal.~~

character_map.h
```diff
@@ -0,0 +1,91 @@
...
+const int CM_HEXADECIMAL_NOT_DECIMAL = 32; // 'a' to 'f' and 'A' to 'F'
+const int CM_DECIMAL = 16 + 8;             // two properties
+const int CM_HEXADECIMAL = 32 + 16 + 8;    // three properties
+
+int ch_map[256];
+
+// must be called early in runtime. Multible calls are ok.
+void initialize_map() {
+
+  // most first characters are non-printable
```
baziotis: Doesn't mean they can't appear in a DIL program. Think of `\n` for example.

character_map.h
```diff
@@ -0,0 +1,91 @@
...
+  ch_map['F'] |= CM_HEXADECIMAL;
+
+  ch_map['#'] = CM_NON_DIL;
+  ch_map['$'] = CM_NON_DIL;
+  ch_map['?'] = CM_NON_DIL;
+  ch_map['@'] = CM_NON_DIL;
+  ch_map['^'] = CM_NON_DIL;
+  ch_map['\\'] = CM_NON_DIL;
+  ch_map['`'] = CM_NON_DIL;
+  ch_map['~'] = CM_NON_DIL;
```
baziotis: We may have to reconsider those. Definitely when we will have strings, these will be legal, but for now, you may remove them altogether.

character_map.h
```diff
@@ -0,0 +1,91 @@
...
+// CM for 'C'haracter 'M'ap. This way we avoid namespaces.
+const int CM_NON_DIL = 0;
+const int CM_DIL_CHAR = 1;                 // character allowed in DIL
+const int CM_BLANK = 2;                    // space, new-line or tab
+const int CM_IDENTIFIER = 4;               // if could be in an ident.
+const int CM_OCTAL = 8;                    // '0' to '7'
+const int CM_DECIMAL_NOT_OCTAL = 16;       // '8' or '9'
+const int CM_HEXADECIMAL_NOT_DECIMAL = 32; // 'a' to 'f' and 'A' to 'F'
+const int CM_DECIMAL = 16 + 8;             // two properties
+const int CM_HEXADECIMAL = 32 + 16 + 8;    // three properties
```
baziotis: Ok, first of all, you don't need to provide properties that given other properties, can be tested. For example, `CM_DECIMAL_NOT_OCTAL` can be tested as `CM_DECIMAL && ~CM_OCTAL`.
Other than that, the usual trick is sort of what you went for with the bitwise operations, taking advantage of powers of 2. To make it more clear, you can do something like:
```cpp
enum {
CM_NON_DIL = 1,
CM_DIL_CHAR = CM_NON_DIL << 1,
CM_BLANK = CM_DIL_CHAR << 1,
...
```
And generally, as always the advice is make the code as simple as possible. We don't need these complicated cases. All these can be tested with simple combinations between simple cases. For example, `CM_DECIMAL` made the code more unpredictable. Below, by looking at the code, it seems that e.g. `6` is octal but not decimal, which is not true.
One has to pretty much do a very good guess that this is handled with this weird addition.

character_map.h
baziotis: As stated above, this code is unintuitive, which unfortunately resulted in wrong interpretation.

character_map_tester.h
```diff
@@ -0,0 +1,20 @@
...
+ */
+#pragma once
+#include "character_map.h"
+
+#include <stdio.h>
+
+// highly technical function
+// to understand the output, check out CharacterMap.h
+void ch_map_test() {
+  char ch[2] = "#";
```
baziotis: Why is that used?

character_map_tester.h
```diff
@@ -0,0 +1,20 @@
...
+ * Authors:    Dimakopoulos Theodoros
+ */
+#pragma once
+#include "character_map.h"
+
+#include <stdio.h>
+
+// highly technical function
+// to understand the output, check out CharacterMap.h
+void ch_map_test() {
```
baziotis: Ok, so you want an automated way to do testing. Printing is not automated because person has to look at the output and verify that it is correct ;) You want a way that a computer checks that the code is correct. That can be done with `assert()`. Do a quick google search if you don't know it, it's very simple.
To use it effectively, you try to find things that _must_ be true when your code runs, e.g. `ch_map['0'] & CM_DECIMAL`, so you `assert` it.

character_map.h
TheodoreAlenas: Woops! Somehow I didn't think of that.

character_map.h
TheodoreAlenas: Shall I collapse them with loops too?

TheodoreAlenas:
Okay!🙂 If it all goes right, the changes will be ready tonight or tomorrow.

character_map.h
```diff
@@ -0,0 +1,91 @@
+/*
+* Copyright:  Copyright Dimakopoulos Theodoros 2019
+* License:    $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+* Authors:    Dimakopoulos Theodoros
+*/
+#pragma once
```
jimisdam, line 6: "#pragma once" isn't the best practice. The ifndef method is more preferred. Like:
```C
#ifndef CHARACTER_MAP_H
#define CHARACTER_MAP_H

// Code...

#endif //CHARACTER_MAP_H```

Comment (baziotis) [reply character_map.h:6]: It's fine. First of all, please when we state that X is not good and Y is better, we should state both why X is not and good _and_ why Y is better. :)
So, basically `#pragma once` is not "standard C++" (i.e. a thing that all C++ compilers have to do the exact same thing) and as all `pragma`s, it depends on how the compiler interprets it. That said, this project does not a hard constraint to be standard C++ when it doesn't have to. In that case, it doesn't have, as any modern compiler treats this as a normal guard.
And we avoid introducing a useless `#define` like `CHARACTER_MAP_H`.

All in all, I don't think there's a right or wrong way in this case, that's why I haven't commented on this in any other PR.

_Note_: I'm glad when other contributors of the project review, but please always have in mind that it's pretty important to explain in depth all the decisions made (or proposed).

**Comment by TheodoreAlenas** [comment]
This is a remake basically, and that is partially why it took quite some time. I hope it's better.

File: character_map.h [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+  // numericals
+  for (int i = '0'; i != '7' + 1; ++i)
+    ch_map[i] |= CM_OCTAL | CM_DECIMAL | CM_HEXADECIMAL;
+
+  for (int i = '8'; i != '9' + 1; ++i)
+    ch_map[i] |= CM_DECIMAL | CM_HEXADECIMAL;
+
+  for (int i = 'a'; i != 'f' + 1; ++i) {
+    ch_map[i] |= CM_HEXADECIMAL;
+    ch_map[toupper(i)] |= CM_HEXADECIMAL;
```
baziotis: Those 3 loops I think are clearer if you write them `for (int i = <start>; i <= <end>; ++i)`

baziotis (commented):
Most of the code is quite good and almost ready for merging. Well done!

TheodoreAlenas:
I hope this does the job.
"most of the code is quite good" means there is plenty of room for changes? If you got the time, could you analyze this a bit? I can do another update if it helps. :-)

character_map.h
```diff
@@ -0,0 +1,80 @@
...
+  CM_NO_PROPERTY = 0,
+  CM_DIL_CHAR = 1,        // character allowed in DIL
+  CM_BLANK = 1 << 2,      // space, new-line or tab
+  CM_IDENTIFIER = 1 << 3, // if could be in an ident
+  CM_OCTAL = 1 << 4,      // '0' to '7'
+  CM_DECIMAL = 1 << 5,    // '0' to '9'
+  CM_HEXADECIMAL = 1 << 6 // '0'-'9', 'a'-'f', 'A'-'F'
+};
+
+int ch_map[128] = {}; //     D E F I N I T I O N
```
baziotis: No need for this comment.

baziotis: Btw, the rest of the code works only if you assume that every element in this array is 0. This _is_ the case, but do you know why?

character_map.h
```diff
@@ -0,0 +1,80 @@
...
+
+  // most first characters are non-printable
+  for (int i = 0; i < 32; ++i)
+    ch_map[i] = CM_NO_PROPERTY;
+  ch_map['\t'] = CM_DIL_CHAR;
+
+  // most characters after 32 are DIL characters
+  for (int i = 33; i < 127; ++i)
+    ch_map[i] = CM_DIL_CHAR;
+  ch_map['\n'] &= ~CM_DIL_CHAR; // this one isn't
```
baziotis: Do me a favor and remove `CM_DIL_CHAR` for now. I keep thinking about it and it seems it will be useless. If we actually need it, we will add it then.

baziotis:
> I hope this does the job.
> "most of the code is quite good" means there is plenty of room for changes? If you got the time, could you analyze this a bit? I can do another update if it helps. :-)

No, it means that "most of the code is quite good" :P . I mean, if something is not good, I will add it in a comment.

baziotis:
My bad... Didn't mean to close it -.-

character_map.h
TheodoreAlenas: Yep. I did it on purpose. It is the ={};

baziotis: Great, but, you could remove that and still be sure that it will be initialized to 0. Why?

TheodoreAlenas: No idea. I have tried such things before in c++ and arrays don't initialize automatically. Will it be?

TheodoreAlenas:
Anything else?

character_map.h
baziotis: The short answer is: Globals are implicitly initialized to 0 if they have no explicit initialization. Generally, this will do it, but:

You can find this in the [C++ standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4713.pdf), paragraph `6.8.3.2 Static initialization`, point 2:
> If constant initialization is not performed, a variable with static storage duration (6.6.4.1) or thread storage duration (6.6.4.2) is zero-initialized.

(We're in the storage duration case, because it is a global but static global duration also have e.g. the `static` variables in a function).

Of course, nobody expects you or me to know the standard inside out, we're not language lawyers. But it's a good reminder that sometimes the standard is useful if you want to be super specific about something.

baziotis:
> Anything else?

I think it's almost ready. 2 things:
1. Generally, as I'm sure you know, we should not put functions in `.h` files (note that I didn't say never though) but in this case, we would have to create another useless file since we don't know yet where this will be used (so, no need to do something, just mentioning it).
2. Please add some tests.

TheodoreAlenas:
@baziotis I better make a cpp file. The code should feel ready when we need it.

baziotis:
> @baziotis I better make a cpp file. The code should feel ready when we need it.

Make a .cpp file for the tests. In this case, I think making a .cpp file just to not have code in a .h file is a bad choice since it's totally trivial to put the initialize in the right place when we know what this place is. I'd prefer it to stay that way but it also doesn't really matter, so do what you like.

TheodoreAlenas:
I ended up adding the cpp file, a test function and a function that prints out what ch_map contains. When I solve some Wi-Fi problems I will upload them. I won't be surprised if the new code needs changes.

baziotis:
> I ended up adding the cpp file, a test function and a function that prints out what ch_map contains. 

The latter one is not needed and not requested. :)
My point here is that my advice for your open source endeavors is to do what is proposed and do it well. This works best for you and other people _both_ when you're the one who receives the reviews or the one who offers them.

> I won't be surprised if the new code needs changes.

Don't worry, it's ok, that's the whole point of this project anyway. But even in huge projects,
the point isn't to not do mistakes. It's to not do the same mistakes ;)

TheodoreAlenas:
Was it right that I made char_map static? I did it because It solved problems that started popping up when making the cpp file. Also, should I call it __char_map?

baziotis:
> Was it right that I made char_map static?

No, for 2 reasons. Well, first a variable should not be in a `.h` file because you will have problems with the compiler (actually, linker). Second, `static` means it is "private" to this file, meaning you can't use it elsewhere. The latter is not a problem since it won't be used anywhere else but the lexer.
For now though, just play with the rules and add them in the `.cpp` file (not `static`) and `extern` them in the tester.
> Also, should I call it __char_map?
No, it's ok for now.

Also, instead of returning the problematic character, just `assert()` what has to be true.

TheodoreAlenas:
@baziotis Okay, so shall I remove the char_map from my file and explain with comments how one can produce it on his own in his file?

Also, seriously... Why are we doing this? C++ has functions, I think, that return wehter a character is blank etc.

baziotis:
> @baziotis Okay, so shall I remove the char_map from my file and explain with comments how one can produce it on his own in his file?

Yes. You just have 3 important files: 1 `.cpp` that contains `initialize_map()` and the map as a global (but not `static`). A `.h` file that `extern`s the map. And a `.cpp` file that has the tests.

> Also, seriously... Why are we doing this? C++ has functions, I think, that return wehter a character is blank etc.

Well, seriously, that's not the best way to request information or ask why you're doing something.
If by C++ you mean the standard library, then yes it has e.g. `isspace()` but not its own function to e.g. test if a character can appear in an identifier. In any case, you could go the way we go or
you could use functions. There's no obvious performance benefit of one over another, it's mostly a matter of preference and not pressing the compiler (because for the fuctions to be effective,
they have to be inlined etc.). Probably the map is a little faster.

TheodoreAlenas:
> Well, seriously, that's not the best way to request information

Sorry! Now that I'm reading it again, you are absolutely right.

character_map_tester.cpp
```diff
@@ -9,24 +9,21 @@
...
-using std::cout;
-using std::endl;
+#include <assert.h>
+ 
 
-int ch_map_problematic_character(int *ch_map) {
+
+void ch_map_tester() {
 
   // sacrificing speed for rigidity.
```
baziotis: No need for that :)

character_map_tester.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright:  Copyright Dimakopoulos Theodoros 2019
+ * License:    $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:    Dimakopoulos Theodoros
+ */
+#pragma once
+// The character map (ch_map) is declared in character_map.h
```
baziotis: No need for that.

character_map_tester.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright:  Copyright Dimakopoulos Theodoros 2019
+ * License:    $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:    Dimakopoulos Theodoros
+ */
+#pragma once
+// The character map (ch_map) is declared in character_map.h
+
+// Exits the probram 
+// if it encoutners a problem with ch_map
```
baziotis: No need for those 2.

baziotis:
> > Well, seriously, that's not the best way to request information
> 
> Sorry! Now that I'm reading it again, you are absolutely right.

It's ok. It's reasonable to ask why we're doing something, it's just better if you phrase it differently.
More importantly, in _this_ project, I try to explain every review I give since I want you to actually understand why we're doing anything.
Is there anything that was not clear why we did it?

Nit: When you address review comments, hit then "Resolve Conversation" to signify that you solved it.

character_map_tester.h
TheodoreAlenas: Are they obvious? Is it usual to call a function "tester" and mean that it can terminate a program? In my intuition, a tester could do anything from changing the character map to signify an error, to using cerr or cout.

TheodoreAlenas:
> Nit: When you address review comments, hit then "Resolve Conversation" to signify that you solved it.

I just tried it on a review comment and it deleted it completely. Did I miss something?

character_map_tester.cpp
TheodoreAlenas: The role of the function changed. It used to return the problematic character. In fact, that is why it's name was so different to the file name

TheodoreAlenas:
> Is there anything that was not clear why we did it?

Everything is clear, no worries. :-)

character_map_tester.h
baziotis: > Are they obvious? Is it usual to call a function "tester" and mean that it can terminate a program?

Yes and yes. It's usual for the tester function to be based on `assert()` that if one fails, then the test fails. The reason that these kinds of tests are helpful is that they're automated. You can just run them and not have to do anything yourself to actually test the code.
There's a whole methodology of development called "test-driven development" that is based
mainly on such kinds of tests.
They're not the only way to test code though and it's good to be aware of that as a lot of people are quite passionate about them. We should test whatever fits best our testing purposes (generally for a compiler, these kinds of tests are very convenient).
You may want to check [this video of Jonathan Blow](https://www.youtube.com/watch?v=21JlBOxgGwY) for a "not-so-popular" opinion.
Last note: These kinds of unit tests are usually called unit tests.

>  In my intuition, a tester could do anything from changing the character map to signify an error, to using cerr or cout.

No, generally a tester function does not change the state of the data although this is not a hard rule. It may use output, or errors, but these kinds of tests are again constructed in way
that they are automated. That is, you don't need to see and verify the output yourself.Usually for these tests, there is an output that _you_ know the program must output (this is the correct output). Then, with the test, you test the output the program yielded against this correct output.

baziotis:
> > Nit: When you address review comments, hit then "Resolve Conversation" to signify that you solved it.
> 
> I just tried it on a review comment and it deleted it completely. Did I miss something?

It probably "folded" the conversation. You can "unfold" it by hitting "Unresolve conversation" which also signifies that "hey, actually this is still not addressed".

character_map_tester.cpp
baziotis: I mean the comment "sacrificing speed for rigidity". That's pretty much assumed in tests :P

TheodoreAlenas:
@baziotis are we close to merging? Any other bits?

TheodoreAlenas:
EMBER RESTORED

And this is one day before I went on vacations!

TheodoreAlenas:
@baziotis seriously though, you did a good job at explaining everything. I learned a lot and I'll be happy to come back! Later. Merry Christmas!

baziotis:
> @baziotis seriously though, you did a good job at explaining everything. I learned a lot and I'll be happy to come back! Later. Merry Christmas!

Thanks mate, merry Christmas to you too! I'm looking forward to seeing a PR from you again.

Issue #10, opened 2019-11-25
Starting the lexer - recognize simple keywords


baziotis:
_Note_: I won't try to replace the theoretical background that compiler textbooks and courses 
provide. Certainly github is not the place for that. But I think it's important to give some
description about what we try to do. Feel free to skip it if you already know the basics.

**What is a lexical analyzer (lexer) ?**
One very important thing that a compiler tries to do is _understand_ the input (so that it can
then tell the computer - in assembly language - what to do based on what the input says).

Now, the input, as it goes into the compiler, is essentially a bunch of characters that (initially)
mean nothing. There has to exist some way to separate / group those characters so that
they have some meaning.

Compare that with natural languages: To understand English text, you first have to separate
the input (i.e. a bunch of characters) into words. Because words are the smallest
piece in the English language that _mean_ something. Then of course, we use _combine_
those  words in different ways to create sentences, paragraphs etc. which help us _convey_ information (i.e. meaning, i.e. help the one reading the text _understand_ what we think).

But it all starts with words, the smallest piece of meaning. Kind of like atoms in the physical world.

So, the same way that a natural language has words, a programming language has tokens.
A token is anything that can mean something on its own. The character `+` is a token on its
own, because it means something on its own. The keyword `int` is also a token.

We want to start off with a lexical analyzer that can recognize simple tokens.
Things like `+`, `int`, `return`, `[` etc.

**How do we implement them?**

There are basically 2 ways to implement lexical analyzers:
1) Automatically generated given a lexical specification.
2) Hand-written lexical analyzers.

In 1), there is a program that takes a lexical specification and constructs a program
that acts as a lexer. This may sounded complicated so feel free to ignore it for now.
We won't use that.

For most programming languages, it's easy enough to make the lexical analyzer by hand.
And the usual way is to make an automata-like program. If you don't know what an automata
means, then again don't worry. Basically, we try to decide based on the characters
that we come across, in what token we might be.
That is, if I see `i`, then the token I'm trying to recognize is definitely _not_ e.g. `return`.
It may be `if` or it may be `int`. So, I try to read a second character that will help me decide etc.

In the general case all those are more difficult because there are user-defined tokens, aka
identifiers.
When you write: `int foo;` you have introduced a user-defined identifier - `foo`.
In the same way, `int i;` introduces the identifier `i`, so when you see `i`, it might be `if`, it might
be `int` but it might be `i`.

[...]

Issue #4, opened 2019-11-25
Provide a function that returns a text version of a token


baziotis:
Interface: `const char *token_name(TOK)`.

Template:
```
const char *token_name(TOK kind) {
    switch(kind) {
    case TOK::EOI:    return "end of input";
    case TOK::INTLIT: return "integer literal";
    ...
    default: assert(0);
    }
    assert(0);
}
```

jimisdam:
Is this supposed to be a typeof thing, like recognizing the type and return it as a string (type's name) ?

baziotis:
No, you pretty much just have to complete the `token_name()` above.

baziotis:
Implemented in: https://github.com/ACM-UoA-Student-Chapter/dil/pull/32

PR #30, opened 2019-12-12
printing keywords

nsoul97:
Issue: #10
If the input string starts with keyword or a special character token (e.g. struct, +), the token is printed. In case we have an identifier, an integer literal or a string that is not a DIL-token nothing is printed.
Keywords(e.g. new) must be followed by a non-identifier character, as opposed to special characters (e.g. &&).

**UPDATE**
The function now uses the Token struct as instructed.


baziotis (commented):
Great PR! Next step is to make a global variable named `token`. This will be of type e.g.:
```cpp
struct Token {
  TOK kind;
};
```
And you will just set the kind instead of printing it out.
Then, we can add tests.

nsoul97:
Should I set the token's kind field to `UNDEFINED` enumeration if no case condition is satisfied instead of printing `"Not a keyword token"`? This would mean that for integer literals and identifiers tokens the global variable token will also have its kind field set to `UNDEFINED`.

baziotis:
> Should I set the token's kind field to `UNDEFINED` enumeration if no case condition is satisfied instead of printing `"Not a keyword token"`? This would mean that for integer literals and identifiers tokens the global variable token will also have its kind field set to `UNDEFINED`.

Yes, for now.

compiler/tests/test_identify_token.cpp
```diff
@@ -0,0 +1,131 @@
...
+  test("while (i<10)", TOK::WHILE);
+  test("false || true", TOK::FALSE);
+  test("true||false", TOK::TRUE);
+
+  test("01210 + 100", TOK::UNDEFINED);
+  test("var_name", TOK::UNDEFINED);
+  test("intvar", TOK::UNDEFINED);
+  test("function", TOK::UNDEFINED);
+  test(" int x;", TOK::UNDEFINED);
+
```
nsoul97, line 55: If the first character is a blank character then kind is set to TOK::UNDEFINED.
This is something I expected from my implementation. Is it ok with what you had in mind for this function?

compiler/tests/test_identify_token.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if (expected_enum != TOK::UNDEFINED)
+    printf("SUCCESS: function identify_token(\"%s\") sets token's kind member "
+           "to %s token.\n",
+           input, token_name(expected_enum));
+  else
+    printf(
+        "SUCCESS: function identify_token(\"%s\") sets token's kind member to "
+        "UNDEFINED token.\n\t \"%s\" starts with an integer literal, an "
+        "identifier, a blank character or a non-DIL string.\n",
+        input, input);
```
baziotis: No need for all that ;)

compiler/tests/test_identify_token.cpp
baziotis: No. For one, the lexer eventually will never put the `UNDEFINED` kind in `token`. If it reaches the point that it would `UNDEFINED`, then it means we have a fatal error and we exit the program.
But for now, think that your function should put the next _available_ kind in `token`. Since there is an available token, it should not put `UNDEFINED` (and of course that means skipping whitespace).

compiler/tests/test_identify_token.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <cstdio>
+
+extern Token token;
+
+/*
+If the global variable token has its kind member set to the expected_enum, the
+function prints a success message. Otherwise, the program execution stops due to
+the assertion.
+*/
+static void test(const char *input, TOK expected_enum) {
```
baziotis: Great idea to create this test function.

compiler/token_name.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "token_name.h"
+#include "tokens.h"
+#include <cassert>
+
+/*
+The function returns the name of the token that corresponds to the enumeration
+which is passed as argument.
+*/
+const char *token_name(TOK kind) {
```
baziotis: Please remove the `token_name*` files because they implement another issue :p

compiler/tests/test_identify_token.cpp
nsoul97: Well at least issue #4 was resolved :p
I actually found the success messages helpful when a test cases went wrong, since I knew the previous successful case. On the contrary, the assert message wasn't really enlightening.

compiler/token_name.cpp
nsoul97: Should I add them with another pr? I checked that they are not assigned to anyone.

compiler/tests/test_identify_token.cpp
baziotis: Yes, because a better assert would show you the call stack so you could see the call path, but well... I get the point. You could implement this function as a macro to solve this issue. But let's stay in this and maybe I'll think of a better solution. Still, implementing 2 issues in the same PR is bad practice, so either implement issue #4 first or don't use it here.

compiler/token_name.cpp
baziotis: Yep.

compiler/identify_token.cpp
```diff
@@ -27,120 +28,140 @@ void init_isIdChar() {
...
+/*
+    The function intializes the isBlankChar array. If a character c:
+    ---> is newline, space or tab character, then isBlankChar[c]=1
+    ---> Otherwise, isBlankChar[c]=0
+*/
+void init_isBlankChar() {
+  isBlankChar[' '] = 1;
+  isBlankChar['\t'] = 1;
+  isBlankChar['\n'] = 1;
+}
```
baziotis: It's not worth it. :) It's just 3 tests worst case with the function potentially inlined by the compiler with the branch predictor doing magic as you will have consecutive white space.

compiler/identify_token.cpp
```diff
@@ -27,120 +28,140 @@ void init_isIdChar() {
...
 found in the input string, if that token is a keyword or a special character.
 If a blank character, an integer literal, an identifier or a non-DIL string is
 encoutered, kind is set to UNDEFINED.
 */
 void identify_token(const char *input) {
-  switch (input[0]) {
+  int ind = 0;
+  while (isBlankChar[input[ind]])
+    ind++;
+  switch (input[ind]) {
```
baziotis: You really strain the compiler for no reason :P Instead of doing `ind++`, do `input++` and the code stays the same as before.

compiler/identify_token.cpp
nsoul97: I used this function to initialize the isBlankChar array, in the same way I initialized the isIdChar array. The information whether a character is blank will be later provided by the character map, so I though I use the information like this for now and just change it later. 
Is the solution you are proposing having a function 
```
bool isBlankChar(char c) {return input==' ' || input=='\t' || input=='\n';}
instead of the isBlankChar array?
```
because the function isBlankChar will be inlined by the compiler, also becuse there are only 3 tests and because the compiler can predict successfully the condition that will be satisfied?

compiler/identify_token.cpp
nsoul97: This is because `input[ind]` is equivalent to `*(input+ind)` right? So I could have `isBlankChar[*input]` instead (1 addition less). One question here is what happens if I have `input[0]` instead of `*input`. Is the addition operation of `input+0` calculated?

compiler/identify_token.cpp
baziotis: > Is the solution you are proposing having a function

Yes. But you're right, you can leave it for now until the character map is ready. It's not that important.

baziotis: > because the function isBlankChar will be inlined by the compiler

Yes

> also becuse there are only 3 tests and because the compiler can predict successfully the condition that will be satisfied?

No, the compiler doesn't predict. It may prove in certain cases. The one who predicts is the CPU.
That was the 3 sentence answer. I will follow with more extended info which anyone who doesn't care can skip.

**Extended Answer**:

-- About branch prediction --
To start simple, with prediction, we mean branch prediction. This is when you have some code like:
```cpp
if (condition) { ... }
```
When the code _runs_, the CPU has something called the branch predictor that for every branch (which is essentially an `if`) tries to predict whether it will be taken or not (i.e. whether you will execute the code inside or you'll skip it. Branches are really bad for modern CPUs (i.e. they slow things down) and so the CPU _predicts_ e.g. that in the next execution of a specific branch, the branch will be taken. So, it then starts executing the instructions "inside" the branch _without yet knowing if the branch should have been taken_ (i.e. the condition is not yet evaluated). Then, at some point, the condition is evaluated, and at this point the CPU may see that the branch should have not been taken. In that case, it has to undo the job it did (because it mistakenly took a branch and started executing its instructions). But, _most_ of the time, the CPU predicts correctly, and so this whole scheme makes the execution a lot faster (because the CPU does not wait for the evaluation of the condition to decide).

A totally naive branch predictor is an 1 bit branch predictor. i.e. if it was taken the last time,
assume that it will be taken now. Then, you can go to a 2-bit, which a little more complex but let's not bore you to death (not way more complex, you can google it).
One important thing to remember though is that the branch predictor in your chip
is _way_ (but I mean _wayyy_) crazier than that. It uses very complex machine learning algorithms and way more bits. You can check [that](https://youtu.be/hgcNM-6wr34?t=831) for a good intro, although you'll have to understand what is pipelining to get it.

-- About the role of the compiler --

Now, all this happens _at runtime_. The compiler can't predict what will happen at runtime (otherwise, we would do symbolic execution). The compiler can only:
1) Prove.
2) Optimize for certain cases based on heuristics.
3) Optimize on certain cases based on runtime.


-1- Prove
Consider this:
```cpp
int test(int a) {
  if (a) return 1;
  return 2;
}

void foo() {
  int v = test(0);
}
```
The compiler here can (potentially) _prove_ (let's skip for a bit how it does it) that you won't take the branch (for this specific call, i.e. test(0)) and so the code is essentially transformed
to: `v = 2`;
But this is not based on the execution, but on something that is constant and known at compile-time.

-2- Optimize for certain cases
The compiler may assume (using heuristics) that e.g. a specific virtual function is called
more frequently than the same (but other) virtual function. So, it may inline it
at the points it is called. This is again, something assumed and something not based
on runtime. Still, these heuristics are studied very much and so this more often than not
is actually an improvement (There's a whole concept related to this called
devirtualization).
You can check [this talk](https://youtu.be/3MRxucTXPdw?t=1352) (in the moment I picked, i.e. 22:30). It is the same concept through function pointers, which is essentially what virtual functions are.

-3- Optimize based on runtime data.
This is something called PGO -> Profile-Guided Optimization, where in simple terms,
the compiler runs your code (having put instructions in it that gather data), gathers
data and then using these data finds that e.g. one function is called way too often so it inlines it.

-- About inlining --
Inling is a misunderstood optimization. Most people think that inlining is beneficial because it eliminates a call and a return to and from a function. That's not really the biggest benefit
in most cases. Inlining is mainly an _enabling_ optimization. That is, it mainly helps
_other_ optimizations. How does it do that? Well, by providing more context.
In the example above, if you inline `test()`, it's way easier to reason that v will take the value 2.
If it wasn't for inlining, you would have to do more sophisticated analysis, analysis that should take consideration code between different functions (there's a whole area for this, called Interprocedural Optimization or IPO).
Now, in this specific case, because your function is so small, the fact that we will eliminate a call and a return instruction will probably be beneficial as well.

I hope that was helpful and not very confusing.

compiler/identify_token.cpp
baziotis: > This is because input[ind] is equivalent to *(input+ind) right? So I could have isBlankChar[*input] instead (1 addition less).

Yes, but consider that the code now has way more additions. While, if you just advance `input` instead of indexing it, the code is essentially the same as before.
Note that the compiler will probably do it for you, but why make its life harder and the readability of the code worse?

> what happens if I have input[0] instead of *input. Is the addition operation of input+0 calculated?

No, it's basically the same thing. Theoretically it should, but the compiler doesn't do it obviously.
I think that in such cases, if you're interested, you can just check the assembly (or IR if you're using LLVM). And ask me if you don't understand something. :)

compiler/identify_token.cpp
nsoul97: That was indeed really helpful. Really great answer.
Some follow up questions:

>This is something called PGO -> Profile-Guided Optimization, where in simple terms,
the compiler runs your code

What if the program when executed expects a user input or some arguments must be be specified from the command line? How will the compiler run the code?

Also in the talk about the function pointers the heuristic optimizations happen based on the profiler, so should't that be considered PGO ?

compiler/tests/test_identify_token.cpp
nsoul97: Perhaps we could create our own `void my_assert(bool)` function that uses the functions
```
int backtrace(void **buffer, int size);
char **backtrace_symbols(void *const *buffer, int size);
```
if the boolean argument is false to print the the call stack information before `assert(0);` or `raise (SIGABRT);`.
If the boolean argument is true nothing happens.
A disadvantage of this solution is that it is OS-specific.

compiler/identify_token.cpp
baziotis: > That was indeed really helpful. Really great answer.

Thanks, I'm glad it was helpful.

>> This is something called PGO -> Profile-Guided Optimization, where in simple terms,
>>  the compiler runs your code

> What if the program when executed expects a user input or some arguments must be be specified from the command line? How will the compiler run the code?

PGO is the idea. Essentially, optimization always implies that you from generic to specific. That is, for optimization to be effective, your data (that your program handles) have to be constrained in _some_ way. The idea is that you test inside this range (this is the very high level idea).

> Also in the talk about the function pointers the heuristic optimizations happen based on the profiler, so should't that be considered PGO ?

Yes, but you could skip that and e.g. imagine that the heuristics transform the code as in the talk based e.g. on how many times the function is called (I mean not on the runtime, but actually in how many places the function is called (which btw is not that simple if you take into consideration loops and recursion)).
For now, what matters is the idea and the idea is that we try to assume some things about the code and then based on that, we do transformations for the most common case (optimization always has to do about optimizing for the most common case).

compiler/tests/test_identify_token.cpp
baziotis: Ok, another kind of long answer. If you implement `my_assert` as a function, then you lose a very important feature - the line number. Note that `assert()` is _not_ a function but a macro. In the macro definition, there is something like `printf("%d", __LINE__);`. This `__LINE__` directive tells the compiler to insert there the line in which you write this `printf`. Because the `assert()` is a define and it gets replaced wherever you call it, you can get the actual line, which is different for every "call" (because it has been replaced).. With a function, the line will always be constant because you always go the same place in source code to execute the function.
So, in your case, you could implement `test` as a macro and basically do achieve what you would like - see what test-case failed based on the line. :)

Other than that, yes implementing backtrace printing is kind of hacky, we probably don't want to that.

baziotis:
It seems to be ready to be merged. I'll leave it a little more because it seems that the character map is almost ready to be merged, so you can then add it here.

baziotis:
Please, refactor this so that it uses the character map. Basically, just remove `character_map.*` files by putting their functionality inside `identify_token.cpp`.

nsoul97:
What about the `compiler/tests/character_map_tester.cpp` file? It includes `character_map_tester.h`, which I cannot find and also there is no `main` function in the file. Probably `ch_map_tester` should be the `main` function and `initialize_map` should be called at the beginning of the function. This is not my test though, so let me know if you want me to change it (in this way or if you have sth else on your mind).

**EDIT**: I actually tried to run also the other tests and all of them(except from the `test_identify_token` which I change in this PR) do not compile because of their inclusions. So perhaps the maintance of the tests should be a seperate issue.

baziotis:
Well, all these will get a good refactoring because now I suppose they don't even compile (and that's no problem, I was not super worried of whether they can compile since they would get a refactoring anyway).
Don't worry about it.

Issue #1, opened 2019-11-25
Read a whole file into memory 


baziotis:
For the compiler, we'll pick a simple and effective scheme to read the input. That is, we will read each file as a whole into memory as a NUL-terminated, contiguous array of characters.

Initially we will have only one input file and we won't support Unicode.

To start off, we can do our job with just a `const char *`. Eventually we might need to make our own `string` class but certainly, it's better to not use `std::string`.

So, we want a function `const char *read_file_into_memory(const char *filepath)`.

ssiatras:
May I work on this one? 

Also, a more general question for the project: Is there preference between high-level C++ commands and low-level system-specific system calls? 

For example on this issue, which one would be better to use?
• istream& get (char* s, streamsize n);
or
• ssize_t read(int fd, void *buf, size_t count);

efsiatras:
May I work on this one? 

Also, a more general question for the project: Is there preference between high-level C++ functions and low-level system-specific system calls? 

For example on this issue, which one would be better to use?
• istream& get (char* s, streamsize n);
or
• ssize_t read(int fd, void *buf, size_t count);

baziotis:
> May I work on this one?

Of course, I'm assigning you. 
> Also, a more general question for the project: Is there preference between high-level C++ functions and low-level system-specific system calls?
> 
> For example on this issue, which one would be better to use?
> • istream& get (char* s, streamsize n);
> or
> • ssize_t read(int fd, void *buf, size_t count);

Whenever there is no significant performance penalty involved or the penalty is not in hot code,
then prefer the high level one because the code is more maintainable.
In this case, even if there was an important penalty with using `istream` (which there isn't),
the code will only run once, in the start. So, unless the penalty is really big, we don't care.
Now, it's important to think of questions like
"What happens with multiple files?", "Is it the best solution to read it all at once?" and maybe
later if the compiler is multi-threaded "how about parallel reading?".
But _all_ these questions are premature optimization and wasted time at this point.

_Note_: Hot code is the parts of a program that have to run fast and usually, multiple times (that's why you will often see the term "hot loop").

efsiatras:
May I use C-style file input/output in order to calculate the size of file properly?

In C++ we have to either read the file char by char or use std::filesystem::file_size of C++ 17.
https://en.cppreference.com/w/cpp/filesystem/file_size

Maybe we need to start a new issue for discussion about the C++ standard version we are going to use eventually.

[...]

PR #31, opened 2019-12-12
Read file

efsiatras:
### Implementation of a function that reads the content of a text file as a whole into memory as a NULL-terminated, contiguous array of characters.

**Issue**: #1 
 
**Params**: filepath = The path of the file to be read.
**Returns**: A pointer to an array of characters (the content of the file).

Things to be noted when this implementation gets updated:

- Various methods for this function:
http://insanecoding.blogspot.com/2011/11/how-to-read-in-file-in-c.html

- About using iterators for this function:
https://stackoverflow.com/a/2602060

- About ftell():
https://stackoverflow.com/a/2076043

- About tellg():
https://stackoverflow.com/a/22986486

- Text vs Binary I/O ('r' vs 'rb' and 'w' vs 'wb'):
http://c-faq.com/stdio/textvsbinary.html
https://stackoverflow.com/a/2174928
https://stackoverflow.com/a/41617906

### DIFF
--- a/compiler/read_file.cpp
+++ b/compiler/read_file.cpp
@@ -1,0 +1,63 @@
+/*
+* Copyright: Copyright ssiatras (Stathis Siatras) 2019.
+* License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+* Authors:   ssiatras (Stathis Siatras)
+*/
+#include "read_file.h"
+#include <cstdio>
+#include <cstdlib>
+
+/*
+ * This function reads the content of a text file as a whole into memory as a
+ * NULL-terminated, contiguous array of characters.
+ *
+ * It only supports ASCII files with one-byte endlines;
+ * (e.g. UNIX format endlines).
+ *
+ * Params:
+ *    filepath = The path of the file to be read.
+ * Returns:
+ *    A pointer to an array of characters (the content of the file).
+ */
+const char *read_file_into_memory(const char *filepath) {
+  if (filepath == NULL) {
+    return NULL;
+  }
+
+  FILE *finput = fopen(filepath, "r");
+  if (finput == NULL) {
+    return NULL;
+  }
+
+  if (fseek(finput, 0, SEEK_END) != 0) {
+    fclose(finput);
+    return NULL;
+  }
+
+  long int fsize = ftell(finput);
+  if (fsize == -1) {
+    fclose(finput);
+    return NULL;
+  }
+
+  if (fseek(finput, 0, SEEK_SET) != 0) {
+    fclose(finput);
+    return NULL;
+  }
+
+  char *string = new char[fsize + 1];
+  if (string == NULL) {
+    fclose(finput);
+    return NULL;
+  }
+  if (fread(string, 1, fsize, finput) != fsize) {
+    return NULL;
+  }
+  string[fsize] = '\0';
+
+  if (fclose(finput) == EOF) {
+    return NULL;
+  }
+
+  return string;
+}
--- a/compiler/read_file.cpp
+++ b/compiler/read_file.cpp
@@ -4,15 +4,13 @@
 * Authors:   ssiatras (Stathis Siatras)
 */
 #include "read_file.h"
-#include <cstdio>
-#include <cstdlib>
+#include <fstream>
+#include <sstream>
+#include <string>
 
 /*
  * This function reads the content of a text file as a whole into memory as a
  * NULL-terminated, contiguous array of characters.
- *
- * It only supports ASCII files with one-byte endlines;
- * (e.g. UNIX format endlines).
  *
  * Params:
  *    filepath = The path of the file to be read.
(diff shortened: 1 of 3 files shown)


baziotis (commented):
Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.

efsiatras:
> Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.

I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.

baziotis:
> > Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.
> 
> I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.

How about `stringstream`?

ssiatras:
> > > Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.
> > 
> > 
> > I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.
> 
> How about `stringstream`?

We would still need a way to know the number of the characters to be read into the stringstream.

efsiatras:
> > > Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.
> > 
> > 
> > I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.
> 
> How about `stringstream`?

Yes, stringstream could also work. It is slightly slower than the current implementation, but it would be much shorter in code. I could implement it, if you want.

baziotis:
> > > > Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.
> > > 
> > > 
> > > I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.
> > 
> > 
> > How about `stringstream`?
> 
> Yes, stringstream could also work. It is slightly slower than the current implementation, but it would be much shorter in code. I could implement it, if you want.

Yes, but we don't care about the performance. ;) And when we don't care about the performance,
we should strive for readability and predictability! In this case, a shorter version helps with that,
so it's better to implement it that way.

ssiatras:
> > > > > Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.
> > > > 
> > > > 
> > > > I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.
> > > 
> > > 
> > > How about `stringstream`?
> > 
> > 
> > Yes, stringstream could also work. It is slightly slower than the current implementation, but it would be much shorter in code. I could implement it, if you want.
> 
> Yes, but we don't care about the performance. ;) And when we don't care about the performance,
> we should strive for readability and predictability! In this case, a shorter version helps with that,
> so it's better to implement it that way.

Done!

efsiatras:
> > > > > Very good PR! Do you think of a different way of implementing `read_file_into_memory` (meaning, shorter way) ? Maybe using some C++ std library features.
> > > > 
> > > > 
> > > > I looked up for every possible way to implement it in a shorter and more C++ style way. The only reliable method I found was using std iterators, which are **much slower** in execution time than the current implementation.
> > > 
> > > 
> > > How about `stringstream`?
> > 
> > 
> > Yes, stringstream could also work. It is slightly slower than the current implementation, but it would be much shorter in code. I could implement it, if you want.
> 
> Yes, but we don't care about the performance. ;) And when we don't care about the performance,
> we should strive for readability and predictability! In this case, a shorter version helps with that,
> so it's better to implement it that way.

Done!

compiler/read_file.cpp
```diff
@@ -0,0 +1,42 @@
...
+  std::ostringstream contents;
+  contents << finput.rdbuf();
+  finput.close();
+
+  std::string str = contents.str();
+  long int fsize = str.size();
+
+  char *cstr = new char[fsize + 1];
+  strcpy(cstr, str.c_str());
+
```
baziotis: You don't need to allocate memory yourself and copy the string as you already have it in `str`. You just have to return `str.c_str()`.

baziotis: However, you then want to _not_ `delete[]` the memory you're getting. Because it's now opaque how the memory is allocated so we don't want to free it (and we don't need to).

efsiatras: Edited:
c_str() returns an array of chars in object from str(). It won't exist after the exit of the function, that's why we have to dynamically allocate it seperately.

baziotis: It could not possibly do that. Well, at least in any sane way. Let's say function `test()` calls `c_str()`. Then, `c_str()` would have to somehow have access to the stack of `test()` in order to allocate an array of chars that would still be valid when `c_str()` returns.

**Note**: That doesn't make sense for various reasons but I'm trying to give a perspective.

efsiatras: I tried to return what c_str() returns. The object string from str() is destructed and its internal memory is cleaned up. That way, the caller of our function receives a dangling pointer.

baziotis: Yes, this is a reasonable concern and it's C++ doing magic. And unfortunately, we have to add a little bit of magic ourselves to prevent the destruction.
Note that this is not a minor detail to just make a copy of the string. The input file might be large, i.e. large enough that 2 of them don't fit into memory.
So, you can create a dummy global variable `std::string` and save there the `contents.str()`. Because it is global, it won't be destroyed when the function exits. The other alternative is to return to your previous solution.
Both alternatives are almost equally bad, so you may want for us to try to think something better (you can do the same too).
For now I propose instead of returning `const char *`, return `std::string`. And just get the contents with `c_str()` in `main()`. It's a reasonable tradeoff.

baziotis: Alternative:
```cpp
size_t get_file_length(std::ifstream &is) {
  is.seekg(0, is.end);
  size_t length = is.tellg();
  is.seekg(0, is.beg);
  return length;
}

std::ifstream is (filepath, std::ifstream::binary);
assert(is);
const char *ret = new char [get_file_length(is)];
is.read(ret, length);
return ret;
```
This is sort of like your first solution but I think it's cleaner.
Pick whatever solution you like of the 2.

efsiatras: If we follow the first solution, should we also change the filepath to string object or keep it as char *?

baziotis: Sorry, I missed that. If I have forgotten a comment / PR / etc. for a long time (say more than 3-4 days), ping me (by e.g. referencing me: @baziotis ). This is also true for other projects that you'll contribute. It's not the easiest thing in the world to keep up with _all_ the activity in a project so people forget stuff or miss them completely.

Now, to your question. No, we don't need to. You can still pass `const char *`.

baziotis:
Ty!

PR #32, opened 2019-12-12
token's text message

nsoul97:
Issue: #4 
Return the token's text message as instructed based on the token enumeration.
Also added a message for the TOK::UNDEFINED enumeration.

### DIFF
--- a/compiler/token_name.cpp
+++ b/compiler/token_name.cpp
@@ -1,0 +1,99 @@
+/*
+ * Copyright: Copyright nsoul97 (Soulounias Nikolaos) 2019.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   nsoul97 (Soulounias Nikolaos)
+ */
+
+#include "token_name.h"
+#include "tokens.h"
+#include <cassert>
+
+/*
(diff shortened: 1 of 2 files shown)


baziotis:
> Issue: #4
> Print the token's text message as instructed based on the token enumeration.
> If the enumeration is TOK::UNDEFINED no message is printed. Instead the program terminates.

It looks good but you don't want it to terminate _here_. This is the job of the lexer, not the printer. ;)

nsoul97:
The function does not print actually. Sorry about that. I actually meant the function returns a string.

About the termination of the program, I actually followed the template of the issue. There is an assert(0) for the default case and not a TOK::UNDEFINED case. Should I add a string to be returned for that case?

baziotis:
> The function does not print actually. Sorry about that. I actually meant the function returns a string.
> 
> About the termination of the program, I actually followed the template of the issue. There is an assert(0) for the default case and not a TOK::UNDEFINED case. Should I add a string to be returned for that case?

Yes, in the `default` we should `assert(0)` because a case that should never happen. The undefined case, because the lexer will never actually finish with an `UNDEFINED` (it will abort), is useless here. But still, we should not pretend that the user knows that. So, yes, add a string (as you can guess, these are all minor details that don't really matter).

